PR #16, opened 2014-07-24
XParseGeometry will always be needed for Windows, no need for cmake to f...

nigels-com:
I see that cmake is now being used to auto-detect things such as this.  However, for non-cmake builds of FreeGLUT needing to needlessly manage NEED_XPARSEGEOMETRY_IMPL seems a bit unreasonable...  Is there  a happy middle ground for this?

### DIFF
--- a/freeglut/freeglut/src/fg_init.c
+++ b/freeglut/freeglut/src/fg_init.c
@@ -317,7 +317,7 @@ void fgDeinitialize( void )
 
 
 /* -- INTERFACE FUNCTIONS -------------------------------------------------- */
-#if defined(NEED_XPARSEGEOMETRY_IMPL)
+#if defined(NEED_XPARSEGEOMETRY_IMPL) || defined(TARGET_HOST_MS_WINDOWS)
 #   include "util/xparsegeometry_repl.h"
 #endif
 


rcmaniac25:
Considering that CMake is pretty much exclusively used, unless you make your own make file, I'm wondering if this is still needed. It's literally been years and I feel both with and without the change work.

nigels-com:
Help me understand this situation better.  A build was broken upstream due to chaanging over to cmake. A simple patch to workaround the most glaring problem is neglected on the basis that cmake should be the only possible way to build FreeGLUT?  There are plenty of situations where cmake is inappropriate, so we'll pretend those don't exist.

rcmaniac25:
Calm down. I have no ability to merge any code. I just happened to go "this has been here for a while. How come nothing has happened with it?" and made my comment. As stated, I feel like both would work. I did not test without CMake as I didn't try to make freeglut without CMake. I asked if this was still needed. Your comment indicates that it is, but I don't know how. The issue description tells me that it doesn't work if you try to build it without CMake and we don't have plain-old makefiles or Autoconf make files. So unless you make your own, completely and 100% possible, makefiles, I wasn't sure how necessary this was.

dcnieho:
Hi Nigel,

Is this the only stumbling block for you? We could merge this indeed. Sorry
it has taken so long. Is this pull the preferred solution for you?

All the best,
Dee

nigels-com:
I can confirm that yes indeed, FreeGLUT is built outside of cmake.  I agree that it's exotic and unusual.  I do think it's unfortunate when cmake becomes the only viable way to build something that isn't all that complicated.

rcmaniac25:
I wouldn't call it exotic. There are so many ways to do the same thing, sometimes you just want the one that works.

PR #24, opened 2014-12-19
Externalized + centralized (some) types visible in freeglut_std.h

sepehr-laal:
Ok, first off, let me start by the fact that C function pointers are extremely pain to make state-aware (well even `glutSetWindowData` is not enough in some cases where callback functions need to be allocated dynamically - imagine being embedded in a project where callbacks are coming from a Lua host, therefore they are not known at the compile time). I wanted not to change the type of the callbacks but I have to.

This is extremely important to C++ users and folks who try to embed Freeglut in their C++ related projects. There needs to be flexibility to change datatypes without breaking the underlying code. 

As a rough example, if I want to change `void(*)(void)` to `std::function<void()>`, everything should still function perfectly fine in theory. This particular example also needs an `allocate` function to initialize the Freeglut's state struct. But you got the idea.

I attempted to change the callback types but everything in Freeglut's main include file (`freeglut_std.h`) is strictly typed in place instead of having a centralized place. So I am proposing the idea of centralizing `typedef`s that are visible via `freeglut_std.h` with this pull request.

### DIFF
--- a/freeglut/freeglut/src/fg_internal.h
+++ b/freeglut/freeglut/src/fg_internal.h
@@ -203,53 +203,7 @@
 
 
 /* -- GLOBAL TYPE DEFINITIONS ---------------------------------------------- */
-
-/* Freeglut callbacks type definitions */
-typedef void (* FGCBDisplay       )( void );
-typedef void (* FGCBReshape       )( int, int );
-typedef void (* FGCBPosition      )( int, int );
-typedef void (* FGCBVisibility    )( int );
-typedef void (* FGCBKeyboard      )( unsigned char, int, int );
-typedef void (* FGCBKeyboardUp    )( unsigned char, int, int );
-typedef void (* FGCBSpecial       )( int, int, int );
-typedef void (* FGCBSpecialUp     )( int, int, int );
-typedef void (* FGCBMouse         )( int, int, int, int );
-typedef void (* FGCBMouseWheel    )( int, int, int, int );
-typedef void (* FGCBMotion        )( int, int );
-typedef void (* FGCBPassive       )( int, int );
-typedef void (* FGCBEntry         )( int );
-typedef void (* FGCBWindowStatus  )( int );
-typedef void (* FGCBJoystick      )( unsigned int, int, int, int );
-typedef void (* FGCBOverlayDisplay)( void );
-typedef void (* FGCBSpaceMotion   )( int, int, int );
-typedef void (* FGCBSpaceRotation )( int, int, int );
-typedef void (* FGCBSpaceButton   )( int, int );
-typedef void (* FGCBDials         )( int, int );
-typedef void (* FGCBButtonBox     )( int, int );
-typedef void (* FGCBTabletMotion  )( int, int );
-typedef void (* FGCBTabletButton  )( int, int, int, int );
-typedef void (* FGCBDestroy       )( void );    /* Used for both window and menu destroy callbacks */
-
-typedef void (* FGCBMultiEntry   )( int, int );
-typedef void (* FGCBMultiButton  )( int, int, int, int, int );
-typedef void (* FGCBMultiMotion  )( int, int, int );
-typedef void (* FGCBMultiPassive )( int, int, int );
-
-typedef void (* FGCBInitContext)();
-typedef void (* FGCBAppStatus)(int);
-
-/* The global callbacks type definitions */
-typedef void (* FGCBIdle          )( void );
-typedef void (* FGCBTimer         )( int );
-typedef void (* FGCBMenuState     )( int );
-typedef void (* FGCBMenuStatus    )( int, int, int );
-
-/* The callback used when creating/using menus */
-typedef void (* FGCBMenu          )( int );
-
-/* The FreeGLUT error/warning handler type definition */
-typedef void (* FGError           ) ( const char *fmt, va_list ap);
-typedef void (* FGWarning         ) ( const char *fmt, va_list ap);
+#include <Gl\freeglut_types.h>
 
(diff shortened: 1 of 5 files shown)


freeglut/freeglut/src/fg_internal.h
```diff
@@ -203,53 +203,7 @@
...
-typedef void (* FGCBMenuState     )( int );
-typedef void (* FGCBMenuStatus    )( int, int, int );
-
-/* The callback used when creating/using menus */
-typedef void (* FGCBMenu          )( int );
-
-/* The FreeGLUT error/warning handler type definition */
-typedef void (* FGError           ) ( const char *fmt, va_list ap);
-typedef void (* FGWarning         ) ( const char *fmt, va_list ap);
+#include <Gl\freeglut_types.h>
```
jtsiomb: how can you fuck this up so badly? wrong case AND wrong slash direction in one include statement....

jtsiomb:
Is there a point to this other than removing duplication? I'm not opposed to removing duplication, but in this case I'm not sure it's worth adding another public header file just for this.

Refrain from using the string "C++" anywhere in your answer. There isn't going to be a C++ compile-time option for freeglut, because it would result in multiple possible versions of binary incompatible libraries, which is unacceptable.

freeglut/freeglut/src/fg_internal.h
sepehr-laal: Dude, you really need to calm down. I gave up on Freeglut because of your
attitude a while ago. This was not even meant to be a final pull request.
It was a work in progress.

Most of these are Visual Studio generated. Windows does not care about case
of paths names and also Visual Studio's compiler does not care about slash
direction.

Not every one is a Linux dev.

You may consider this pull request abandoned.

On Tuesday, May 5, 2015, John Tsiombikas <notifications@github.com> wrote:

> In freeglut/freeglut/src/fg_internal.h
> <https://github.com/dcnieho/FreeGLUT/pull/24#discussion_r29728084>:
>
> > -typedef void (* FGCBInitContext)();
> > -typedef void (* FGCBAppStatus)(int);
> > -
> > -/* The global callbacks type definitions */
> > -typedef void (* FGCBIdle          )( void );
> > -typedef void (* FGCBTimer         )( int );
> > -typedef void (* FGCBMenuState     )( int );
> > -typedef void (* FGCBMenuStatus    )( int, int, int );
> > -
> > -/* The callback used when creating/using menus */
> > -typedef void (* FGCBMenu          )( int );
> > -
> > -/* The FreeGLUT error/warning handler type definition */
> > -typedef void (* FGError           ) ( const char *fmt, va_list ap);
> > -typedef void (* FGWarning         ) ( const char *fmt, va_list ap);
> > +#include <Gl\freeglut_types.h>
>
> how can you fuck this up so badly? wrong case AND wrong slash direction in
> one include statement....
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dcnieho/FreeGLUT/pull/24/files#r29728084>.
>


-- 

*Sepehr Laal* /* C++ developer, Downstream, Portland, OR */
[](auto o, auto e) {
  o << "*Mobile* " << "*+1* *(541) 556 6064*" << e;
  o << "*Desk*   " << "*+1* *(503) 496 7295*" << e;
}(std::cout, std::endl);

nigels-com: Aside from the broader potential of C++-ification of the GLUT API, and adding yet another public header, I'd raise a separate concern:

It's fine for FreeGLUT to internally use a FGCB prefix for callback types.  But I think promoting these from implementation to the formal interface is problematic.  It's effectively extending the GLUT API in a way that is incompatible with GLUT - actually proliferating these types into application code will result in GLUT-incompatible applications.

I personally think that C++ support ought to be wrapped around FreeGLUT (or GLUT) rather than this more intrusive approach.

- Nigel

jtsiomb: On Tue, May 05, 2015 at 05:59:13PM -0700, Sepehr Laal wrote:
> Dude, you really need to calm down. I gave up on Freeglut because of your
> attitude a while ago. 

Suit yourself.

> Most of these are Visual Studio generated. Windows does not care about case
> of paths names and also Visual Studio's compiler does not care about slash
> direction.

Is that an excuse for crappy unportable code in a cross-platform
project? Don't use visual studio's automatic tools if they can't emit
correct code, or at the very least check the output before sending pull
requests. Nobody owes you the effort to fix your code. Do it properly or
don't do it at all.

> Not every one is a Linux dev.

Not everyone has to be, to write correct include statements. It just
takes half a clue.

> You may consider this pull request abandoned.

Suit yourself.

-- 
John Tsiombikas
http://nuclear.mutantstargoat.com/

jtsiomb: On Tue, May 05, 2015 at 06:12:42PM -0700, Nigel Stewart wrote:

> Aside from the broader potential of C++-ification of the GLUT API, and
> adding yet another public header, I'd raise a separate concern:
> 
> It's fine for FreeGLUT to internally use a FGCB prefix for callback
> types.  But I think promoting these from implementation to the formal
> interface is problematic.  It's effectively extending the GLUT API in
> a way that is incompatible with GLUT - actually proliferating these
> types into application code will result in GLUT-incompatible
> applications.
> 
> I personally think that C++ support ought to be wrapped around
> FreeGLUT (or GLUT) rather than this more intrusive approach.

Indeed, I agree 100% with all of the above. You put words to my nagging
doubts about this change.

Cheers.

-- 
John Tsiombikas
http://nuclear.mutantstargoat.com/

PR #29, opened 2015-03-26
Implement initial Wayland support

Tarnyko:
This commit creates a new "wayland" backend, which can be
activated at configure time with "-DFREEGLUT_WAYLAND=ON".

If done so, it will be used instead of X11 (building both
and doing runtime detection may become possible later).
Please note that if you choose to use GL instead of GLES
(by not specifying "-DFREEGLUT_GLES=ON"), then libX11
will still be pulled as an indirect dependency.

Following features are still WIP :
- menus (not implemented, TODO) ;
- client-side decorations (not implemented, required
  because Wayland shells do not draw title bars nor
  resize grips, TODO) ;
- game mode (code is commented out, depends on WIP
  protocols tested upstream, WAIT FOR UPSTREAM) ;
- window visibility states (code is commented out,
  depends on xdg-shell protocol, TODO).

Signed-off-by: Manuel Bachmann <tarnyko@tarnyko.net>

### DIFF
--- a/freeglut/freeglut/CMakeLists.txt
+++ b/freeglut/freeglut/CMakeLists.txt
@@ -52,8 +52,10 @@ ENDIF()
 # OpenGL ES support
 OPTION(FREEGLUT_GLES "Use OpenGL ES (requires EGL)" OFF)
 
-# option to build either as "glut" (ON) or "freeglut" (OFF)
 IF(NOT WIN32)
+    # Wayland support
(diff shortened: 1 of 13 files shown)


Tarnyko:
Please find a demonstration video (best watched in fullscreen mode) here :

 https://www.youtube.com/watch?v=FQvRIoALQyg

Tarnyko:
Closing this pull request, to recreate a new one :).

PR #31, opened 2015-07-10
Fix compilation on Linux.

jzarl:
Commit ce15044f7362943aee7d465bf20310ba02991dae introduced an error
"undefined reference to `sball_initialized'" on non-Windows operating
systems.

### DIFF
--- a/freeglut/freeglut/src/fg_spaceball.c
+++ b/freeglut/freeglut/src/fg_spaceball.c
@@ -11,7 +11,7 @@
 #include <GL/freeglut.h>
 #include "fg_internal.h"
 
-#if(_WIN32_WINNT >= 0x0501)
+#if( !_WIN32 || _WIN32_WINNT >= 0x0501)
 
 /* -- PRIVATE FUNCTIONS --------------------------------------------------- */
 


dcnieho:
@jtsiomb, Does this look good to you?

jtsiomb:
It looks good in the sense that: yes, this apparently was broken, and yes, this change fixes it.

However, I don't think it's the correct way to fix this. Now that I look at that piece of code, I find that ifdef there completely wrong to begin with. It's pretending to care about windows versions, yet there is not a single piece of windows-specific code in that file (which is good, because it's not under mswin/ ).

I'm violently opposed to bringing back random system-specific ifdefs in the non-system-specific files. It goes against all that effort a couple of years ago of separating the system-specifics in subdirectories.

I assume that this ifdef can be moved in mswin/fg_spaceball_mswin.c and make fgPlatformInitializeSpaceball() and friends conditionally empty within that file, instead of not having them at all, and having to avoid calling them from the platform-independent part with conditionally empty fgInitializeSpaceball() and friends.

I hope that part was clear, I'm a bit sleep-deprived at the moment, just returned from a trip.

dcnieho:
Hi John,

That did make some sense, thanks. Lets see what the author of the windows spaceball code has to say about it. I'll pull in the fix for now, but we have to keep this open to fix it properly.

@stonexjr Could you have a look at this and perhaps implement what John suggests?

Thanks,
Dee

PR #32, opened 2015-07-10
Cmake package

jzarl:
This branch creates a FreeGLUTConfig.cmake (and accompanying files), needed for the find_package config mode in cmake.

There already is the FindGLUT.cmake find package module in the cmake distribution, but it is unmaintained and technically is too late to do proper detection of GLUT. E.g. FindGLUT can only speculate about required dependencies based on the current platform, whereas a proper cmake module can set definitions such as "-DFREEGLUT_GLES" correctly.


dcnieho:
This looks good. I'll just merge it while others may have further comments on it down the road. Thanks!

rcmaniac25:
Finally got around to looking at these. All is good, I tested it (BlackBerry 10 and Windows) and it worked, which was very cool.

I ran into one issue: When building for Windows, it looks up `FREEGLUT_LIB_PRAGMAS` and tries to use pragmas for loading *.lib(s). This tends to be fine if paths are correct, but the `FreeGLUTTargets.cmake` file doesn't specify `FREEGLUT_STATIC` for `FreeGLUT::freeglut_static`, so it fails when trying to find the lib. If I explicitly state the define or change the library to `FreeGLUT::freeglut` then it compiles fine. So I'd recommend adding the `FREEGLUT_STATIC` define for the static library.

Sample project I used had the following CMakeLists.txt:
```
CMAKE_MINIMUM_REQUIRED(VERSION 2.6.0)
PROJECT(FreeGLUTTest)

SET(FG_CONFIG_ROOT "${CMAKE_INSTALL_PREFIX}")
IF (WIN32)
	IF (NOT EXISTS CMAKE_INSTALL_PREFIX)
		SET(FG_CONFIG_ROOT "${CMAKE_INSTALL_PREFIX}/../freeglut")
	ENDIF()
ENDIF(WIN32)

INCLUDE(${FG_CONFIG_ROOT}/lib/cmake/FreeGLUT/FreeGLUTConfig.cmake)

ADD_EXECUTABLE(FreeGLUTTest
	src/main.c
)

TARGET_LINK_LIBRARIES(FreeGLUTTest
	PUBLIC FreeGLUT::freeglut_static
)
```

jzarl:
I don't have a windows machine nearby ATM, so I could not test this. Could you try applying ffc24cfb76 and report whether this is an appropriate fix for you?

rcmaniac25:
You're quick. That did the trick and it works now.

PR #41, opened 2016-01-08
Emit the correct LIBNAME to .pc files

mingwandroid:
Irrespective of "FREEGLUT_REPLACE_GLUT" being set to "No" or being on
Win32 where this variable does not even exist, the .pc file contained
Libs: -L${libdir} -lglut

Instead, emit the correct value.


dcnieho:
thanks!

PR #42, opened 2016-01-17
User data callback support

rcmaniac25:
Added additional functions (end in "Ucall") that support user data.

Fixed potential issue where, when determining OS, Android, BlackBerry, and Wayland were overlooked.

See discussion thread for "[State-aware callbacks](http://sourceforge.net/p/freeglut/mailman/freeglut-developer/thread/BLU179-W571F31D61384D8697CE6B0D0710%40phx.gbl/)" on Freeglut Developer Mailing list

### DIFF
--- a/freeglut/freeglut/src/fg_callback_macros.h
+++ b/freeglut/freeglut/src/fg_callback_macros.h
@@ -34,9 +34,7 @@
  */
 
 /*
- * Compiler defines:
- * FG_COMPILER_SUPPORTS_GCC_VA_ARGS_HACK: if the compiler supports GCC's variadic macro implementation (AKA, ##__VA_ARGS__)
- * FG_COMPILER_SUPPORTS_VA_ARGS: if the compiler supports variadic macros
+ * Compiler define: FG_COMPILER_SUPPORTS_VA_ARGS: if the compiler supports variadic macros
  */
 
 /* What supports variadic macros based off Wikipedia article on it (GCC-like must support C99 or higher to use variadic macros) */
@@ -49,15 +47,6 @@
 #define FG_COMPILER_SUPPORTS_VA_ARGS 1
 #else
 #define FG_COMPILER_SUPPORTS_VA_ARGS 0
-#endif
-
-/* If __VA_ARGS__ is supported, it needs to be GCC(-like) or Clang (since Clang mimics GCC) */
-#if FG_COMPILER_SUPPORTS_VA_ARGS && \
-	(defined(__GNUC__)) || \
-	(defined(__clang__))
-#define FG_COMPILER_SUPPORTS_GCC_VA_ARGS_HACK 1
-#else
-#define FG_COMPILER_SUPPORTS_GCC_VA_ARGS_HACK 0
 #endif
 
 /*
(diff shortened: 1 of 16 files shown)


freeglut/freeglut/src/fg_internal.h
```diff
@@ -526,11 +579,14 @@ typedef void (*SFG_Proc)();
  * and for no other reason.  Since it's hidden in the macro, the
  * ugliness is felt to be rather benign.
  */
-#define SET_WCB(window,cbname,func)                            \
+#define SET_WCB(window,cbname,func,udata)                      \
 do                                                             \
 {                                                              \
     if( FETCH_WCB( window, cbname ) != (SFG_Proc)(func) )      \
```
dcnieho: what if user wants to change udata but not func?

freeglut/freeglut/src/fg_internal.h
```diff
@@ -548,6 +604,47 @@ do                                                             \
     ((window).CallBacks[WCB_ ## cbname])
 
 /*
+ * FETCH_USER_DATA_WCB() is used as:
```
dcnieho: there is a second slightly different version of this comment right below, accident?

freeglut/freeglut/src/fg_internal.h
```diff
@@ -572,9 +670,10 @@ do                                            \
 {                                             \
```
dcnieho: not related to what you did, but any idea why there would be two definitions of INVOKE_WCB? Second is more compact but presumably first works on all platforms? I don't see a difference expect first is split up a bit more for legibility?

freeglut/freeglut/src/fg_menu.c
```diff
@@ -780,14 +780,29 @@ void fghCalculateMenuBoxSize( void )
...
+    FREEGLUT_EXIT_IF_NOT_INITIALISED ( "glutCreateMenuUcall" );
     if (fgState.ActiveMenus)
         fgError("Menu manipulation not allowed while menus in use.");
 
-    return fgCreateMenu( callback )->ID;
+    return fgCreateMenu( callback, userData )->ID;
+}
+
+/* Standard glutCreateMenu */
+static void glutCreateMenuCallback( int menu, FGCBUserData userData )
```
dcnieho: so if the original callback setter is called (e.g. glutCreateMenu in this case), you put an internal helper as the callback function and the user provided callback as the userData, and then internally handle calling the user callback without user data. Do i understand that correctly (and you use the same pattern in other places)?

- if this is an internal function, start it with fg or fgh (h for helper), not glut. all glutXXX() are exposed to users.
- can we generalize this helper, implement it once, and give it a better name showing its routing function (or what to call that)?

freeglut/freeglut/src/fg_internal.h
rcmaniac25: I don't know why there are two versions of this, but I did try to remove the Windows only version and Visual Studio 2015 failed to compile... so there's something it doesn't like with the second macro.

If you only want the first macro, that would probably work on all platforms (untested). Assuming there isn't someone currently around who wrote the second macro to try and have the compiler inline the function call, without a register being taken up by `func`. If that was the design goal, I'd be amazed if "now", that has any measurable performance difference.

freeglut/freeglut/src/fg_internal.h
rcmaniac25: Woops, type. I'll fix that along with any other comments that I implement.

freeglut/freeglut/src/fg_internal.h
rcmaniac25: I didn't see a reason why the developer couldn't re-provide the function (nearly all window callbacks are just set functions anyway), nor a reason why the developer couldn't provide a struct pointer of some sort that could have it's contents modified.

I tried to keep all the functions consistent, and had written the window callbacks last. I didn't see a good place for many other callback functions to have an additional set operation, and since window callback data is per-callback, it could result in a bunch of `glut` functions who's implementation would all be:
```
void glutSetXXXUserData(void* user_data)
{
    SET_WCB_USER_DATA(XXX, user_data);
}
```

Not to say it can't be done, but I'm not sure why I should pollute the `glut` namespace more when the callbacks can be re-set to the window with new data.

I'll take the opportunity to note that near the end of my initial implementation of the window callbacks, I noticed an already implemented user data parameter for windows and menus (`glutSetWindowData`/`glutGetWindowData`, `glutSetMenuData`/`glutGetMenuData`). It was per-window/per-menu, as opposed to this per-callback version, and seems to be a freeglut implementation as opposed to Glut. At the time I had a "oh come on, someone already did this?!" but given they were broad and weren't in "the unwritten standard for callback user data" format, I questioned how many people knew about it and used it. While I probably can't use it within the implementation I provided, I'm not sure if anything should be done with it or if it should be left as is. For all I know, someone will say "skip your implementation and just do that". Understandable, but devs (myself included) tend to look straight ahead and not look what other functions exist unless it's a new library. Suddenly adding callbacks that take user data will be noticed, even when something similar may have existed for a decade or more. So will more of the same get people to use it?

freeglut/freeglut/src/fg_menu.c
rcmaniac25: Correct, that is the format I use for all callbacks so that old and new functions continue to work.

* Ah, I see what you're getting it. Yea, I can change the `glutXXXCallback` functions to be `fghXXXCallback`. Didn't think about that (and didn't know the difference between `fg` and `fgh` as I didnt' see it mentioned anywhere).
* I did generalize this for fg_callbacks, but it's a bit ugly and for the functions that don't use it, there are some spots where it can't be generalized. I was able to do it for window callbacks because all of them (except for keyboard callbacks) only use ints. I'll look into it and see what I can do.

freeglut/freeglut/src/fg_internal.h
dcnieho: Maybe we should bring this up on the forum, separate from your current work. Indeed any modern compiler should optimize this difference away with ease.

freeglut/freeglut/src/fg_menu.c
dcnieho: Thanks! Seems like a sensible approach to make old and new work.
- whether function should be `fg` or `fgh` doesn't really, matter i don't think its consistent in the current codebase anyway. but some prefix like that shows its internal, not API.
- if generalizing becomes such a mess that there is no benefit, then it may not be worth it. Your call!

freeglut/freeglut/src/fg_internal.h
dcnieho: Sorry, i wasn't clear: i was referring to line 585, `if( FETCH_WCB( window, cbname ) != (SFG_Proc)(func) )`. If you call the setter again, only changing userdata, it'll be ignored. I don't see the need for separate setters either, that'd be hell.

I'm not sure how much the existing user data places are used. Guess on the user side you could have it point to a switchyard where you then select the right member function (if you're using userdata to enable using freeglut with member functions), this new implementation would be easier. You will see the docs for whats already there think of very simple scenarios only, not making living with freeglut from C++ easy. (I find, e.g., http://stackoverflow.com/questions/14865339/use-non-member-method-as-a-callback-in-glut, not easy and not flexible enough when you have more than one class to work with for instance. As such, i am for this addition. But perhaps it is best to see what others on the list say before going further.

Personally i also never read whole APIs, but i do search if i need something, and i would hope these functions would come up.

rcmaniac25: Ah. Good eye. I can either add an additional check in the if statement or have a separate else if after that one which only sets user data. Opinions?

Rarely do read a whole API. Certain ones I do... then I keep the header and when a new version comes out, I diff it so I know what changed.

freeglut/freeglut/src/fg_internal.h
rcmaniac25: I'll double check for sanity sake, but I'm pretty sure VS 2015 failed.

freeglut/freeglut/src/fg_menu.c
rcmaniac25: I'll see with the generalization. It may just require some refactoring.

freeglut/freeglut/src/fg_internal.h
dcnieho: Its not a strong opinion, but i guess else if makes logical sense / keeps code easiest to understand

rcmaniac25: Done

freeglut/freeglut/src/fg_internal.h
rcmaniac25: Ok, VS2015 compiles it fine. I don't know about older compilers. I'm writing an email for this right now.

freeglut/freeglut/src/fg_internal.h
dcnieho: Looks good to me! Thanks

rcmaniac25:
I just thought of a potential bug. I'm going to try to produce it and a fix (if needed) later in the day.

rcmaniac25:
Updated the docs to mention user-data callbacks

rcmaniac25:
Made callbacks use generic generators if possible.

dcnieho:
Nice work again man! Let me try and ping the list again, I'd like more eyes on this

rcmaniac25:
No problem

rcmaniac25:
So... if it wasn't noticed, I didn't get around to this. I should spend the hour or two to demo this. Any opinions on a demo to show? (I'll think of something if nobody responds by the time I get to this...)

dcnieho:
Hi Vinnie, thanks a lot for your continued interest in this.
As for the demo, its a bit hard to see what to write indeed. A shining example would be to enable callbacks to member functions of C++ classes. But then we'd be adding C++ source to a C project, which would mean a special (default off i guess) flag in the CMake file for this particular demo to be built, etc. Can you think of a use case for this in pure C?

rcmaniac25:
Well, got a demo finally. Unless something else is desired, I believe this is finally done.

Note: I did think about a function to release user data, but then I realized:

1. It over complicates it
2. We don't provide memory allocator access (maybe we should?) and control all our own memory. This is the reverse of that, they provide the memory, they free the memory.
3. This would add a whole new set of functions needed to set these callbacks and require either storing those callbacks or some other "setup" that, as per 1), would over-complicate things.

freeglut/freeglut/src/fg_callback_macros.h
```diff
@@ -0,0 +1,412 @@
...
+ * all. Stuff like MSVC6... It works by having a specific-count macro that "extracts" each argument to have them reused without the
+ * parathesis.
+ *
+ * A note on parathesis, as earlier mentioned, if the GCC variadic macro element was standard, then instead of needing:
+ *
+ * func EXPAND_WCB(Mouse)(( (GLUT_LEFT_BUTTON, GLUT_DOWN, 10, 30), userData));
+ *
+ * ...you can do the following:
+ *
+ * func EXPAND_WCB (GLUT_LEFT_BUTTON, GLUT_DOWN, 10, 30);
```
dcnieho: needs ", userData"?

freeglut/freeglut/src/fg_internal.h
```diff
@@ -215,52 +217,97 @@
...
+typedef void (* FGCBInitContextUC   )( FGCBUserData );
+typedef void (* FGCBAppStatus       )( int );
+typedef void (* FGCBAppStatusUC     )( int, FGCBUserData );
 
 /* The global callbacks type definitions */
-typedef void (* FGCBIdle          )( void );
-typedef void (* FGCBTimer         )( int );
-typedef void (* FGCBMenuState     )( int );
-typedef void (* FGCBMenuStatus    )( int, int, int );
+typedef void (* FGCBIdle            )( void ); \
```
dcnieho: "\"?

freeglut/web-src/docs/api.php
```diff
@@ -327,6 +327,40 @@
 when you created it.</ul>
 </p>
 
+<h3>3.2.2 User-data callbacks</h3>
+
+<p>
+GLUT was created as a tool to help teach OpenGL programming. To simplify
+development, callbacks were used for handling display, input, and other
+events. But the time it was developed, the purpose, or for some other
```
dcnieho: "but at the time"

dcnieho:
Nice work! All demos work for me, and i think the new one demos it nicely. Even in a simple C program, what you demonstrate tastes significantly better than the style the original timer demo was forced to adopt at that time. 

See a few tiny comments on the diff

As for releasing user data, maybe i am seeing this too simple, but that isn't our problem at all? We just return a memory address (pointer), and its up to the user to make sure this remains valid (its his data after all) and to know how to interpret this at all. Or am i missing anything? :)

Ok, last hurdle is to, one last time, bring this up on the mailing list. Its a significant change and some testing and diff reading by others wouldn't hurt. Wont leave it hanging too long tho. Let me know when you're ready to take this step. In fact, feel free to write the list as its all your honor :)

freeglut/freeglut/src/fg_callback_macros.h
rcmaniac25: Nope. Look at the comment below this line

rcmaniac25:
I'll address the couple comments.

As for data validity: yep, it's up to the dev. If we are to maintain it, then we need memory handling callbacks or similar. We simply provide the user their pointer.

I intend to bring it up with the mailing list, just haven't gotten to it yet.

freeglut/freeglut/src/fg_callback_macros.h
rcmaniac25: To be more explicit, the expand function automatically adds ", userData" to the arguments

rcmaniac25: And... screwed up both comments. So, since GCC variadic macro extension is not standard, a workaround must be written. This is explaining how the code would look different if it was a standard.

jtsiomb: Let's not leave an ifdef-ed piece of dead code, just to demonstrate the beauty of a non-standard compiler feature :)

PR #45, opened 2016-01-18
API documentation and macros updates

stonexjr:
Updated API documentation for glutSpaceballXXX functions.
Updated comments for macros WCB_spaceXXX in fg_internal.h
Added new macros for Spaceball buttons.


rcmaniac25:
My only worry is that freeglut_std.h is usually only for functions/macros that were part of the original GLUT. I'd move those macros to freeglut_ext.h

stonexjr:
Done.

dcnieho:
Thanks!! The button constants are probably good to have, to avoid magic constant. But as its not in original GLUT, it would be good to add what they map to. And that they can be ORed together if multiple buttons are down (right?).
Looking great!

stonexjr:
@dcnieho The values of the new macros are set such that they are ready to be ORed in the case when multiple buttons are pressed. But current implementation does not consider this case. It is treated as two separate press events. Further, I've noticed that the [dwKeystate](https://github.com/stonexjr/FreeGLUT/blob/Jinrong/freeglut/freeglut/src/mswin/fg_spaceball_mswin.c#L184) can have multiple bits set, if multiple buttons are pressed down. But unlike keyboard event WM_CHAR, keeping pressing down the button(s) does not keep generating WM_INPUT events, so the multiple-button-pressed-down event can only be detected at the moment when the second button is pressed while the first button is already being hold pressed.
So, the first argument of glutSpaceButton callback only indicates which button has just changed state (pressed or released). I would probably sleep with that, unless the callback can have a third argument to tell how many buttons in total are pressed down.

dcnieho:
@stonexjr Thats for explaining that! It's not exactly clear to me from what you said whether its always single keys that are given through the first argument of glutSpaceButton callback, even if one key is already pressed down and a second is just depressed. If so, great! (i suppose that's compatible, right?)

But is the numerical value of the first argument given equal to what previous glut would give you (you just put a convenient name on there)? I think the docs would be enhanced by noting the numerical value each of these five possible values resolve to. Also, as we've tried to do for other functions in the docs, it would be good to have a "changes from GLUT" section, noting that these GLUT_SPACEBALL_BUTTON_A are not in original glut. You can pretty much copy-paste that from another bit of the manual?

Do you see a chance to make these changes? Then I'm more than happy with your work and will merge it right away. thanks!

stonexjr:
Yes, a single key is always given by the first argument of glutSpaceButton callback.

Yes, the numerical value of the first argument is equal to what previous glut would give me.

dcnieho:
cool, will put it in tonight. thanks!

PR #48, opened 2016-07-08
Added FREEGLUT_REPLACE_GLUT configure option for MS Windows

tehnick:
It provides back compatibility with older (autotools based) freeglut versions. Also it simplifies build rules for crossplatform programs: they may be linked with -lglut regardless of the system and library type (shared/static).

Patches were originally prepared for MXE project: https://github.com/mxe/mxe/pull/1399, https://github.com/mxe/mxe/pull/1429


freeglut/freeglut/CMakeLists.txt
```diff
@@ -391,12 +396,22 @@ ENDIF()
...
+
     LIST(APPEND LIBS winmm)
     IF(FREEGLUT_BUILD_SHARED_LIBS)
         TARGET_COMPILE_DEFINITIONS(freeglut PRIVATE FREEGLUT_EXPORTS)
+        SET_TARGET_PROPERTIES(freeglut PROPERTIES OUTPUT_NAME ${LIBNAME})
     ENDIF()
     IF(FREEGLUT_BUILD_STATIC_LIBS)
         TARGET_COMPILE_DEFINITIONS(freeglut_static PUBLIC FREEGLUT_STATIC)
+        IF(FREEGLUT_REPLACE_GLUT AND NOT FREEGLUT_BUILD_SHARED_LIBS)
+            SET_TARGET_PROPERTIES(freeglut_static PROPERTIES OUTPUT_NAME ${LIBNAME})
```
dcnieho: Thanks for this! One question here. If i read this correctly, this causes the static build to be called glut, in the case only static build is requested. Is that a valid way to replace glut (i.e., won't lead to trouble)?

tehnick: As far as I remember MinGW in some cases generates both *.dll  and *.a files. And in this case *.a file is not a static library but additional file for proper linking with shared library. So I decided to add this check to avoid file name collisions. But I have not tested build of both (shared+static) libraries simultaneously, so I am not sure if this check is really necessary. I'll test it now.

tehnick: Hmm, the additional file is *.dll.a, so this check is useless.

tehnick:
Updated.

tehnick:
Now it looks like this:
```
$ tree ./x86_64-w64-mingw32.static-out/
./x86_64-w64-mingw32.static-out/
├── build.log
└── usr
    ├── bin
    │   └── libglut.dll
    ├── include
    │   └── GL
    │       ├── freeglut_ext.h
    │       ├── freeglut.h
    │       ├── freeglut_std.h
    │       └── glut.h
    └── lib
        ├── cmake
        │   └── FreeGLUT
        │       ├── FreeGLUTConfig.cmake
        │       ├── FreeGLUTConfigVersion.cmake
        │       ├── FreeGLUTTargets.cmake
        │       └── FreeGLUTTargets-release.cmake
        ├── libglut.a
        ├── libglut.dll.a
        └── pkgconfig
            └── glut.pc

8 directories, 13 files
```

tehnick:
And in such case (when both libraries are available) MinGW prefers shared library for linking.

tehnick:
Hi. Are any other issues with this pull request?

rcmaniac25:
It looks good, my only question is why you need the else statement (https://github.com/dcnieho/FreeGLUT/pull/48/files#diff-598f53e0e3d1e89d0626616d2706392aR401) which otherwise is already set by default?

dcnieho:
Hi Vinnie,

I see no use in that else either. I guess i should merge after taking that
out?

Boris, can you remove lines 401 and 402 and update the pull request?

All the best,
Dee

tehnick:
Sorry that I completely forgot about this PR.

> I see no use in that else either. I guess i should merge after taking that
out?
> Boris, can you remove lines 401 and 402 and update the pull request?

Yes, sure. Done.

tehnick:
Any comments?

rcmaniac25:
You took care of my comments, but I don't have merge access. @dcnieho can you merge if you have no other issues?

dcnieho:
Sorry for the delay, thanks a lot for sending in this pull request!

PR #53, opened 2017-04-25
NetBSD build fix and some cleanups

0-wiz-0:


dcnieho:
Hi, thanks for the contributions. I grepped the code, and see that both the hatmap variables and the static functions are used in fg_joystick_x11.c. If this is platform (X11) specific code, perhaps it should be moved to that file. @jtsiomb?

jtsiomb:
I'm not very familiar with the joystick code. I'll have a look as soon as possible and get back at you. In general if it's something only used by the X11 code, then sure makes sense to put it in the X11-specific file to avoid ifdefs or dead code.

0-wiz-0:
Thanks for the review, @dcnieho.
I guess I should put back the hatmap's (or move them), but static functions can not be used from a different compilation unit. Or am I missing something?

dcnieho:
Sorry for the very long delay. It seems like the removed variables and functions are used in \src\x11\fg_joystick_x11.c. A few years back, we didn't have these platform-specific files. Something must have gone wrong when splitting up the large files we had, and in this case never have been picked up because the functions and variables are only used when #ifdef HAVE_USB_JS. I guess thats not the case for many builds, so its never been found. 

I'll merge your other commits manually and then close this. I'll move the code over myself. Thanks a lot for your contribution!

dcnieho:
Ok, done. @0-wiz-0, would you mind trying the current master and see if things build as they should and you do not get warnings about this dead code?

0-wiz-0:
@dcnieho Thanks for merging.

Current master does _not_ build on NetBSD, there are the following problems:

`freeglut/freeglut/src/x11/fg_joystick_x11.c` is missing the header inclusion/define stuff from `freeglut/freeglut/src/fg_joystick.c`:
````
diff --git a/freeglut/freeglut/src/x11/fg_joystick_x11.c b/freeglut/freeglut/src/x11/fg_joystick_x11.c
index 30ad4b5..c8bf278 100644
--- a/freeglut/freeglut/src/x11/fg_joystick_x11.c
+++ b/freeglut/freeglut/src/x11/fg_joystick_x11.c
@@ -41,6 +41,63 @@

 #include <fcntl.h>

+#if defined(__NetBSD__) || defined(__FreeBSD__) || defined(__FreeBSD_kernel__)
+
+#    ifdef HAVE_USB_JS
+#        if defined(__NetBSD__)
+#            ifdef HAVE_USBHID_H
+#                include <usbhid.h>
+#            else
+#                include <usb.h>
+#            endif
+#            include <dev/usb/usb.h>
+#        elif defined(__FreeBSD__) || defined(__FreeBSD_kernel__)
+#            ifdef HAVE_USBHID_H
+#                include <usbhid.h>
+#            else
+#                include <libusbhid.h>
+#            endif
+#            include <legacy/dev/usb/usb.h>
+#        endif
+#        include <dev/usb/usbhid.h>
+
+/* Compatibility with older usb.h revisions */
+#        if !defined(USB_MAX_DEVNAMES) && defined(MAXDEVNAMES)
+#            define USB_MAX_DEVNAMES MAXDEVNAMES
+#        endif
+#    endif
+
+struct os_specific_s {
+  char             fname [128 ];
+  int              fd;
+  int              is_analog;
+  /* The following structure members are specific to analog joysticks */
+  struct joystick  ajs;
+#    ifdef HAVE_USB_JS
+  /* The following structure members are specific to USB joysticks */
+  struct hid_item *hids;
+  int              hid_dlen;
+  int              hid_offset;
+  char            *hid_data_buf;
+  int              axes_usage [ _JS_MAX_AXES ];
+#    endif
+  /* We keep button and axes state ourselves, as they might not be updated
+   * on every read of a USB device
+   */
+  int              cache_buttons;
+  float            cache_axes [ _JS_MAX_AXES ];
+};
+
+/* Idents lower than USB_IDENT_OFFSET are for analog joysticks. */
+#    define USB_IDENT_OFFSET    2
+
+#    define USBDEV "/dev/usb"
+#    define UHIDDEV "/dev/uhid"
+#    define AJSDEV "/dev/joy"
+
+
+#endif
+
````

When I fix that, there are compilation warnings about unused variables:
````
.../freeglut/freeglut/src/x11/fg_joystick_x11.c:104:12: warning: ‘hatmap_x’ defined but not used [-Wunused-variable]
 static int hatmap_x[9] = {0, 0, 1, 1, 1, 0, -1, -1, -1};
            ^
.../freeglut/freeglut/src/x11/fg_joystick_x11.c:105:12: warning: ‘hatmap_y’ defined but not used [-Wunused-variable]
 static int hatmap_y[9] = {0, 1, 1, 0, -1, -1, -1, 0, 1};
            ^
.../freeglut/freeglut/src/x11/fg_joystick_x11.c:139:12: warning: ‘fghJoystickFindUSBdev’ defined but not used [-Wunused-function]
 static int fghJoystickFindUSBdev(char *name, char *out, int outlen)
            ^
.../freeglut/freeglut/src/x11/fg_joystick_x11.c:165:12: warning: ‘fghJoystickInitializeHID’ defined but not used [-Wunused-function]
 static int fghJoystickInitializeHID(struct os_specific_s *os,
            ^
````

and then a linking error, because it does not pull in libusbhid (-lusbhid):
````
[ 32%] Linking C executable bin/timer_static
lib/libglut.a(fg_joystick_x11.c.o): In function `fghJoystickInitializeHID':
fg_joystick_x11.c:(.text+0x308): undefined reference to `hid_get_report_desc'
fg_joystick_x11.c:(.text+0x3ca): undefined reference to `hid_report_size'
fg_joystick_x11.c:(.text+0x41b): undefined reference to `hid_start_parse'
fg_joystick_x11.c:(.text+0x6a2): undefined reference to `hid_get_item'
fg_joystick_x11.c:(.text+0x6b6): undefined reference to `hid_end_parse'
*** Error code 1
````

Can you please fix these?

dcnieho:
Thanks for testing! Should i move that whole bunch of defines over to `/x11/fg_joystick_x11.c`? or are they also needed in `/fg_joystick_x11.c`? I'd like to keep that clean.

Would you be able to make a new pull request fixing these things? That'd be more efficient than me making changes and asking you to test.
Those variables may well be undefined due to the very ifdefs in which they are used. Perhaos their declaration should have a more specific ifdef around it as well

dcnieho:
hmm, better use a new pull request for that, the branches have diverged now. closing again...

PR #57, opened 2017-06-14
CMake error on X11 builds if XInput.h is not found

rcmaniac25:


freeglut/freeglut/CMakeLists.txt
```diff
@@ -324,6 +324,8 @@ IF(UNIX AND NOT(ANDROID OR BLACKBERRY OR FREEGLUT_WAYLAND))
         # Needed for multi-touch:
         CHECK_INCLUDE_FILES("${X11_Xinput_INCLUDE_PATH}/X11/extensions/XInput2.h" HAVE_X11_EXTENSIONS_XINPUT2_H)
         LIST(APPEND LIBS ${X11_Xinput_LIB})
+    ELSE()
+        MESSAGE(FATAL_ERROR "Missing X11's XInput.h (X11/extensions/XInput.h)")
```
dcnieho: Should we add 'install xinput development package, e.g., "apt-get install libxi-dev"'? Just as a helpful hint

rcmaniac25: I thought about it... but assumed that the package name is not the same on every platform. So didn't want to fix it to one platform. (such as with `apt-get`)

dcnieho: Ok, fair enough. If you need the help, such potentially wrong help can only lead you further astray.

PR #62, opened 2017-11-14
Unicode

ccxvii:


freeglut/freeglut/src/mswin/fg_main_mswin.c
```diff
@@ -480,6 +480,81 @@ fg_time_t fgPlatformSystemTime ( void )
...
+			HANDLE object = GetClipboardData(CF_UNICODETEXT);
+			if (object)
+			{
+				WCHAR *wtext = GlobalLock(object);
+				if (wtext)
+				{
+					int n = WideCharToMultiByte(CP_UTF8, 0, wtext, -1, NULL, 0, NULL, NULL);
+					if (n > 0)
+					{
+						char *text = malloc(n);
```
rcmaniac25: I cannot find a `free` function on this and other buffers when freeglut shuts down.

freeglut/freeglut/src/mswin/fg_main_mswin.c
```diff
@@ -1023,6 +1104,7 @@ LRESULT CALLBACK fgPlatformWindowProc( HWND hWnd, UINT uMsg, WPARAM wParam, LPAR
         break;
 #endif
 
+#if _MSC_VER > 1400
```
rcmaniac25: Perhaps make this `#ifdef WM_SETCURSOR` as I don't believe this has changed for multiple versions, and IIRC, this built pre-Visual Studio 2005... so I don't know what changed, but under that assumption, I worry that two VS 2005 or pre-2005 builds could be different.

freeglut/freeglut/include/GL/freeglut_ext.h
```diff
@@ -157,6 +157,7 @@ FGAPI void    FGAPIENTRY glutSetMenuFont( int menuID, void* font );
 /*
  * Window-specific callback functions, see fg_callbacks.c
  */
+FGAPI void    FGAPIENTRY glutKeyboardExtFunc( void (* callback)( int, int, int ) );
```
rcmaniac25: Should we have a `glutKeyboardUpExtFunc`? For completeness?

freeglut/freeglut/src/fg_callbacks.c
```diff
@@ -126,6 +126,7 @@ IMPLEMENT_GLUT_CALLBACK_FUNC_ARG0_2NAME(MenuDestroy, Destroy)
 
 /* Implement all these callback setter functions... */
 IMPLEMENT_CURRENT_WINDOW_CALLBACK_FUNC_ARG2(Position)
+IMPLEMENT_CURRENT_WINDOW_CALLBACK_FUNC_ARG3_USER(KeyboardExt,int,int,int)
```
rcmaniac25: Please add the public function for the `glutKeyboardExtFuncUcall` to `freeglut_ucall.h`

freeglut/freeglut/src/mswin/fg_main_mswin.c
```diff
@@ -686,7 +766,8 @@ static LRESULT fghWindowProcKeyPress(SFG_Window *window, UINT uMsg, GLboolean ke
             if( ToAscii( (UINT)wParam, 0, state, code, 0 ) == 1 )
                 wParam=code[ 0 ];
 
-            INVOKE_WCB( *window, KeyboardUp,
+            if (wParam < 128)
```
rcmaniac25: Will this break existing GLUT functionality? I don't think I've tried Unicode keys so I don't know how this shows up currently.

freeglut/freeglut/src/mswin/fg_main_mswin.c
```diff
@@ -1375,10 +1458,13 @@ LRESULT CALLBACK fgPlatformWindowProc( HWND hWnd, UINT uMsg, WPARAM wParam, LPAR
             break;
 
         fgState.Modifiers = fgPlatformGetModifiers( );
-        INVOKE_WCB( *window, Keyboard,
-                    ( (char)wParam,
-                      window->State.MouseX, window->State.MouseY )
+        INVOKE_WCB( *window, KeyboardExt,
+                ( wParam, window->State.MouseX, window->State.MouseY )
         );
+        if (wParam < 128)
```
rcmaniac25: Same as other comment:
Will this break existing GLUT functionality? I don't think I've tried Unicode keys so I don't know how this shows up currently.

freeglut/freeglut/src/x11/fg_main_x11.c
```diff
@@ -579,6 +579,285 @@ __fg_unused static void fghPrintEvent( XEvent *event )
     }
 }
 
+/* UTF-8 decoding routine */
```
rcmaniac25: This should probably get moved up to `fg_misc.c` or `fg_main.c` as it will might be needed for Android and Wayland and isn't X11 specific.

freeglut/freeglut/src/x11/fg_main_x11.c
```diff
@@ -934,46 +1227,71 @@ void fgPlatformProcessSingleEvent ( void )
...
-                        fgState.Modifiers = INVALID_MODIFIERS;
+                        i += chartorune(&c, utf8 + i);
+
+                        /* ...one for the Unicode translateable keypresses... */
+                        if (keyboard_ext_cb)
+                            keyboard_ext_cb(c, event.xkey.x, event.xkey.y, keyboard_ext_ud);
+
+                        /* ...one for the ASCII translateable keypresses... */
+                        if (keyboard_cb)
+                            if (c < 128)
```
rcmaniac25: Same comment as before: Will this break existing GLUT functionality?

I get it would only be one part of the whole Unicode, but I don't know who might be relying on that value.

rcmaniac25:
Thanks for that work. I think Unicode has long been missing, and the clipboard is a nice addition.

freeglut/freeglut/include/GL/freeglut_ext.h
ccxvii: A character may be the result of several keypresses, such as the compose key or extended latin "dead key" input, and especially so when typing for example chinese text using an IME. As such, there is no corresponding 'up' event for a character event coming from the system. In windows the characters come from a WM_CHAR event, and on X11 the XFilterEvents function removes events eaten by the IME.

freeglut/freeglut/src/mswin/fg_main_mswin.c
ccxvii: It might break existing programs, but only if they were taking the signed character and casting it to an unsigned value and using it as latin-1. I'm fine with removing this check and maintaining the old 'bad' behavior if you feel it is needed for API stability.

rcmaniac25: Hmm... I might wait to see if someone else has any opinions.

freeglut/freeglut/include/GL/freeglut_ext.h
rcmaniac25: Ok. Good to know.

jtsiomb: Which is why it's confusing to call this glutKeyboardExtFunc. A better name might be glutCharFunc similar to the glfw notation.

freeglut/freeglut/src/mswin/fg_main_mswin.c
jtsiomb: Yes I'm afraid this has the potential to break existing programs, and should be avoided.

freeglut/freeglut/include/GL/freeglut_ext.h
rcmaniac25: If we're suggesting different names, why not be obvious: `glutKeyboardUnicodeFunc`?

ccxvii: I picked the name because it was similar to the original behavior of glut where glutKeyboardFunc was used to detect characters, and glutSpecialFunc was used to detect presses of non-character keys. This was before glutKeyboardUpFunc was added to support games wanting to know key press events, and conflated the event of a keyboard key going down and up and the event of a character being generated.

As such, it may be better to call the new function glutCharacterFunc or something similar, to make clear that distinction... unfortunately glutKeyboardFunc applies the keyboard mapping and sends events on a key down event, with some special hacks to do the same mapping to KeyboardUp events. Backwards compatibility prevents us from changing the behavior to what I would have done instead: glutRawKeyDownFunc, glutRawKeyUpFunc, which take scan codes and make glutKeyboardFunc take a 32-bit unicode character rather than a 7-bit ascii character.

rcmaniac25: Then perhaps glutUnicodeFunc?

jtsiomb: I agree, Char or Character in the name might be lead someone to think that it's another interface for the same thing. I had the same confusion while reading the glfw docs. I still don't like glutUnicodeFunc for some reason, but I'll aggree it's the best candidate so far. I'll post again if I can thing something better.

freeglut/freeglut/include/GL/freeglut_ext.h
```diff
@@ -175,6 +175,15 @@ FGAPI void    FGAPIENTRY glutSetWindowData(void* data);
...
 
+/*
+ * Clipboard access
+ */
+#define GLUT_CLIPBOARD 0
+#define GLUT_PRIMARY 1
+#define GLUT_SECONDARY 2
+FGAPI void         FGAPIENTRY glutSetClipboard(int selection, const char *string);
+FGAPI const char * FGAPIENTRY glutGetClipboard(int selection);
+
```
jtsiomb: I don't see anything wrong with this API, but I think every new API we add should be discussed in the mailing list first, because we'll be stuck with it for ever. This github mirror is not the standard place to take decisions about public interfaces.

Could you please post in the freeglut-developers mailing list with your proposal for the new APIs? It's better, and more direct, if you do it.

freeglut/freeglut/include/GL/freeglut_ext.h
dcnieho: Either glutUnicodeFunc or glutUnicodeCharFunc would work best for me i think, don't see other options. And this is higher level than the keyboard in our current scheme of things

jtsiomb:
I think it's fair to say that after all this time, this specific pull request as it stands will not be merged. I reviewed again the changes proposed here, and the short resulting discussion in the mailing list from back then, and there are a few problems which held this back all this time as I see it, even though it has merit, and would make a useful addition to freeglut. In order of importance:

  1. Changing existing freeglut callbacks in a backwards-incompatible way and more importantly incompatible with original GLUT, is out of the question. If the concept of using ASCII characters for keypress/release callbacks was flawed to begin with, we're stuck with it. The only way to "fix" this, is to introduce entirely new API entry points.
  2. Untangling the mess of how to deal with the existing API and any new API additions, deciding how it should all work, and making sure we're not breaking anything, and not creating new problems to be carried on for ever, seems to be more work than anyone cared to put into it. There were a couple of messages in the list, and it just petered out. Deciding on new API additions that need to be carried over for ever is tiring, and requires at least one person with the motivation to see it through and drag the rest of us on :)
  3. This pull request and the proposal in the mailing list tries to do too many things at once: making sure keydown/keyup events match up (possibly by using some kind of virtual scancodes, see mailing list), making it possible to use get unicode input, and for some reason also adding some kind of clipboard support. At the very least the last item should be a separate pull request.

About the clipboard support, which is the easier part of this to merge if it was a standalone pull request, I considered for a while whether I should ask the original author to separate that bit out in a new pull request and just merge it. It's clearly something useful. But I'm not entirely sure it falls in the bounds of what GLUT is supposed to be: a simple library to write cross-platform OpenGL programs. I don't know ... I could be convinced either way.

About the unicode input. I could see this going forward, if there is a strong will by someone to do all the legwork, propose a reasonable new set of callbacks that maps nicely across all major supported systems (windows, UNIX/X11). Show how the proposal does not affect backwards compatibility, and make a convincing argument about how it's the best way to do it, and it won't be another faulty API we'll have to maintain for the rest of time. This should not be a pull request, but discussion. In a github issue or the mailing list.

In all cases, every new feature should be in a small self-contained pull request, for it to have any chance of being merged.

Regardless, I want to thank the original author of this pull request for the effort, and everyone involved in the discussions.

PR #68, opened 2018-10-19
Fix and enable -fvisibility=hidden by default when building with gcc

ldv-alt:
This short series fixes build with gcc -fvisibility=hidden, and enables -fvisibility=hidden by default when building with gcc.  As result, 168 internal symbols are no longer exported by the shared library.


rcmaniac25:
So the one question I see is that you mark the font definitions in X11 l with FGAPI which, if I understand, will not be exported in the shared library. So how would a program link against them with GCC using X11?

rcmaniac25:
I also just realized, you only do this for GCC. Is it not possible for Clang or whatever non-Windows compiler is used?

ldv-alt:
Just the otherwise, I mark these font variables with FGAPI to make them exported.

dcnieho:
@tsiomb, should this be merged?

jtsiomb:
I don't see any harm in hiding any symbols which aren't part of the API. The only potential issue, is breaking programs which use some internal freeglut symbol to augment the functionality. For example using, `fgDisplay` to gain access to our connection to the X server. I looked at the list of symbols that will stop being visible, and I can't think of anything else someone might want to use, so I'd say just adding an FGAPI to `fgDisplay` just to avoid any such breakage.

Also in the list of symbols that will become unaccessible with this patch, I see a bunch of glut-prefixed joystick functions. I assume these should have the FGAPI prefix too.

dcnieho:
Ok, shall i merge and add FGAPI to fgDisplay?

ldv-alt:
On Mon, Sep 02, 2019 at 10:27:01PM -0700, Diederick C. Niehorster wrote:
> Ok, shall i merge and add FGAPI to fgDisplay?

The patch has been tested in Sisyphus which means that fgDisplay is not needed to be exported in GNU/Linux.

I don't mind if FGAPI is added to fgDisplay if you suppose it might be needed elsewhere.

jtsiomb:
> I don't mind if FGAPI is added to fgDisplay if you suppose it might be needed elsewhere.

Yes, I'd say let's play it safe and add FGAPI to fgDisplay, and also to the joystick functions which for some reason don't already have it. Otherwise I'm fine with merging this.

jtsiomb:
I see this was never merged. But I'm also having second thoughts about whether this should be merged.

I'm slightly nervous about having the public `FGAPI` definition in `freeglut_std.h` changed to add the visibility attribute. This should probably be something that affects only freeglut's build, not user programs. Maybe it should be hidden behind something like the windows `FREEGLUT_EXPORTS` macro used to select between dllexport and dllimport.

I'm also not sure if the `#ifdef __GNUC__` block is guaranteed to be sufficient for the default visibility attribute to be added for all compilers where cmake decided to hide the symbols based on the `CMAKE_C_VISIBILITY_PRESET` setting.

kusma:
For the visibility thing, using [__has_attribute](https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html) is probably the right approach.

jtsiomb:
This pull request will not be merged as is, and has not been updated in years, so I'm closing it.

If anyone cares enough about the visibility issue to restart the process of merging something, I suggest opening an issue first, or sending an email to the freeglut-developer mailing list. Such a proposal should discuss the approach and explain how it doesn't affect the public API/ABI, and how it interacts with build settings controlling visibility.

PR #69, opened 2018-11-18
BeforeClose

ghost:
On Windows platform "BeforeClose" callback function is called to confirm window closing.

### DIFF
--- a/freeglut/freeglut/src/mswin/fg_main_mswin.c
+++ b/freeglut/freeglut/src/mswin/fg_main_mswin.c
@@ -780,7 +780,6 @@ LRESULT CALLBACK fgPlatformWindowProc( HWND hWnd, UINT uMsg, WPARAM wParam, LPAR
     SFG_Window *window;
     LRESULT lRet = 1;
     static int setCaptureActive = 0;
-    int doClose = TRUE;
 
     FREEGLUT_INTERNAL_ERROR_EXIT_IF_NOT_INITIALISED ( "Event Handler" ) ;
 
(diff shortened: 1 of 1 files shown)


rcmaniac25:
While I think it looks good, I'd probably say "can you add for X11"

berge2007:
BeforeClose calling added to X11 but not tested.
I'm working on Windows so please somebody who works with X11 test the code.

For test need a simple callback function:
`void OnBeforeClose(int *doClose)
{
	*doClose = FALSE;
}` 
and set the callback function:
`glutBeforeCloseFunc(OnBeforeClose);`

I've added BeforeClose callback function calling code when WM_SYSCOMMAND/SC_CLOSE message received to Windows specific source code

rcmaniac25:
@jtsiomb @dcnieho this looks fine to me. Any other opinions?

jtsiomb:
First of all I might have missed something, but I only see a bunch of changes here, without any justification on why they should be merged. Merging them just because the code "works", especially since they are adding a new public API entry point, is not sufficient justification. What are you trying to solve by adding a new entry point for "before closing" a window? Is it really necessary?

BTW these discussions should be carried in the project mailing list. Not here.

Also, from a quick look at the X11-specific code. It seems that the WM_DELETE_WINDOW ICCCM procool is used to detect when the window is about to close, but this event will only be received when the window manager tries to close the window (when the user clicks some window decorations button for instance), not if it's closed through some mechanism, like if the application itself exits or closes the window. Is that the intention and equivalent to the proposed windows functionality? Without any justification about this whole thing, and a clear statement of what it's supposed to achieve I can't tell.

berge2007:
If the user close a FreeGLUT window, FreeGLUT destroys the window instantly and without mercy :) 
My intention was to add an API point to FreeGLUT to take control over closing mechanism. So programmer (for example) could make a  "Do you want to save your work?" pop up and depend on the user answer maybe the closing process should be aborted or let it be executed, or maybe just signaling an other mechanism  

IMHO the modification is thin and well targeted, and it won't break backward compatibility nor degrade the performance nor make any structural changes in FreeGLUT.

jtsiomb:
For future reference (and because I had no recollection of this and had to go digging in the mailing list archives), this is pretty much where this discussion stopped in seeming agreement, but went no further: https://sourceforge.net/p/freeglut/mailman/message/36485204/

PR #71, opened 2019-03-18
Remove unnecessary libs from linking

ChristianS99:
https://bugs.gentoo.org/658936

It seems, ${X11_LIBRARIES} pulls in some libraries, that aren't actually needed.
As far as I could see, only libX11 is actually needed.
The change to the CMakeLists.txt should fix this.


jtsiomb:
This needless dependency was probably introduced with the switch from autotools to cmake, because X11_LIBRARIES seemed like the correct thing to use. Other than libX11 we're only conditionally using extension libraries like Xrandr, Xf86vm, Xinput etc, which however as far as I can tell are handled separately in the cmake file anyway.

Let's merge this and I'll test it when it hits the svn repo. If it breaks something I'll fix it there, by adding the fine-grained cmake directives instead of the catch-all X11_LIBRARIES.

ChristianS99:
I took a closer look at those, and noticed that the CMake variables used are not mentioned in documentation (https://cmake.org/cmake/help/latest/module/FindX11.html). I would guess, that FindX11 was changed with new names. The names used in freeglut CMakeLists.txt are still set, just not documented.

I'm just not sure if HAVE_X11_EXTENSIONS_XF86VMODE_H and HAVE_X11_EXTENSIONS_XINPUT2_H should be changed to follow the names from FindX11 module, as I'm not sure, what they are for.

jtsiomb:
The problem with cmake, is that it changes way too much and too often.

For better or worse at the moment we require version 3.0.0 according to the first line of the CMakeLists.txt (which might even be needlessly high). Looking at the documentation for that version instead of the latest version which you have linked, you'll see that the original variables are the correct ones to use.

So for example in cmake 3.0.0 the variable for checking the presence of the XFree86 VidMode extension is X11_xf86vmode_FOUND and not X11_xf86vm_FOUND. Please revert this last change before your pull request can be merged.

The HAVE_X11_EXTENSIONS_*_H things are defines we're setting, to be checked in various places in the code, and they have nothing to do with cmake, so they definitely shouldn't be changed either.

ChristianS99:
You are right. I didn't pay attention to the cmake version that is required.
Reverted it.

jtsiomb:
Commited, svn revision 1845. Seems to work fine. Thank you.

rcmaniac25:
@dcnieho can you update the Github mirror when you get the chance?

dcnieho:
Done!

Croydon:
Fixed by https://github.com/dcnieho/FreeGLUT/commit/f6459749021057de40e98d6eea2dd0c95d56acdc

PR #75, opened 2019-10-21
Use <sys/joystick.h> on netbsd, too.

coypoop:
<machine/joystick.h> is a compatibility header that only includes
<sys/joystick.h>, and wasn't added to newer architectures.

This helps the build on netbsd/aarch64.

### DIFF
--- a/freeglut/freeglut/src/wayland/fg_internal_wl.h
+++ b/freeglut/freeglut/src/wayland/fg_internal_wl.h
@@ -105,18 +105,7 @@ struct tagSFG_PlatformWindowState
 /* XXX The below hack is done until freeglut's autoconf is updated. */
 #        define HAVE_USB_JS    1
 
-#        if defined(__FreeBSD__) || defined(__FreeBSD_kernel__)
-#            include <sys/joystick.h>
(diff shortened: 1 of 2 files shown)


dcnieho:
Thanks! As i am not familiar with this platform: would this change entail any possible backward compatibility problems?
@jtsiomb, this look ok to you?

jtsiomb:
I don't have experience with NetBSD. If there's a way to detect NetBSD version, and we know which version added the `sys/joystick.h` header, ideally we could add some logic to use one or the other.

Failing that, I'd say merge it if the build is broken on current NetBSD systems, and we'll deal with any backwards compatibility issues when and if they crop up.

coypoop:
sys/joystick.h has been a thing on netbsd since 2004, so I don't expect it to be necessary to do a version check

jtsiomb:
That doesn't mean there shouldn't be a check. It just makes it acceptable to not bother until someone comes along who is affected by that and fixes it :)

I have no objection to merging this.

dcnieho:
Thanks both for the discussion. merging...

PR #76, opened 2020-01-21
fg_gl2: fix build failure against gcc-10

trofi:
On gcc-10 (and gcc-9 -fno-common) build fails as:

```
[ 31%] Linking C executable bin/Lorenz_static
/usr/bin/cmake -E cmake_link_script CMakeFiles/Lorenz_static.dir/link.txt --verbose=1
/usr/lib/ccache/bin/cc  -Wall -pedantic  -rdynamic CMakeFiles/Lorenz_static.dir/progs/demos/Lorenz/lorenz.c.o \
   -o bin/Lorenz_static  -lGLU -lGL -lm -lX11 -lXrandr -lXxf86vm -lXi -lm lib/libglut.a -lGL -lm -lX11 -lXrandr -lXxf86vm -lXi
ld: lib/libglut.a(fg_gl2.c.o):(.bss+0x0):
  multiple definition of `fghGenBuffers'; lib/libglut.a(fg_window.c.o):(.bss+0x0): first defined here
ld: lib/libglut.a(fg_gl2.c.o):(.bss+0x8):
  multiple definition of `fghDeleteBuffers'; lib/libglut.a(fg_window.c.o):(.bss+0x8): first defined here
```

Note: duplicate definition comes here from 'fg_gl2.h'.
The fix is to move declaration to a 'fg_gl2.c' file.

gcc-10 will change the default from -fcommon to fno-common:
https://gcc.gnu.org/PR85678.

The error also happens if CFLAGS=-fno-common passed explicitly.

Bug: https://bugs.gentoo.org/705840


dcnieho:
This looks good to me, but I'm unable to test the GL2 stuff. Can someone test this please before merge?

jtsiomb:
Responding to the original message in the automatically linked bug report which brought me here:

trofi said:
> have accidental redundant definitions where declarations only are expected (openrc, freeglut, many others)

There's nothing accidental about this. It's perfectly valid in C, not valid in C++. The new defaults of gcc-10 are broken, and we may have to work around this, but that doesn't make the current code wrong.

dcnieho:
I don't see anything wrong with this pull request. At least I can say it compiles here, and I don't see anything in it that will affect the runtime functionality of the code to require any further testing.

I don't like the duplication of this whole block of definitions in two different places (one with extern one without). I prefer the classic C way of declaring them once in the header file and letting the linker deal with it; I think it's much more elegant that way. But I have no real objection to merging this in order to work around the gcc-10 issue.

The other way would be to add `-fcommon` to the compiler flags, which is what I did in my other projects, but I don't want to mess with cmake to do that. It might involve detecting the compiler to add the flag only for gcc, in case other compilers don't know about this flag, and I certainly don't want to test multiple compilers to make sure. So this duplication might be the easiest and most future-proof way to fix the problem.

mattst88:
Can we merge this?

jtsiomb:
> Can we merge this?

If you're anxious to get something fixed, take the time to figure out the proper place to report it. This is *not* the freeglut project site. It's a mirror, and if you want something merged *here* you'll have to wait for dcnieho to see this, and decide to merge it in his mirror.

I will go on and fix this in the official freeglut repo, because I see that it affects increasing number of people. But that I even saw this thread at all is pure luck, because someone else pointed it out to me in an issue of an unrelated project of mine.

For future reference, the proper place is either the freeglut bug tracking *on sourceforge*, or the freeglut-developer mailing list. See http://freeglut.sourceforge.net for details.

mattst88:
Thanks!

> If you're anxious to get something fixed, take the time to figure out the proper place to report it.

I'm aware. I mailed your mailing list a few months ago requesting that you finally make a working release. No need to be condescending.

Perhaps this github repo is doing more harm than good, if you don't want contributions here.

> Edit: done, svn revision 1863. Don't have gcc 10. Test and report back.

You don't need it:

> The error also happens if CFLAGS=-fno-common passed explicitly.

jtsiomb:
> I should also mention that your mailing list never accepted my mail. If you want people to mail your list, it would be good if it accepted mail from non-subscribed addresses.

No, you need to subscribe first. Otherwise just post a bug report on sourceforge. You also need to subscribe to sourceforge to do that, so whichever you find more convenient of the two.

> Perhaps this github repo is doing more harm than good, if you don't want contributions here.

It's not a matter of "we" not wanting contributors here. "We", collectively, have nothing to do with this repo. It belongs to one of the three freeglut maintainers, and he maintains this mirror because he finds it's more convenient for himself, and for others others to contribute here.

But it is still a non-official mirror, and when you contribute here you are relying on a single person to find the time to deal with your contribution. It's just a practical matter: if you need to get something fixed fast, you post it to the official place where everyone involved in the project can potentially see it, and do something about it. Or at least you try here first if you find it more convenient, and if you don't get a response try the official venue to increase your chances.

> You don't need it (gcc-10), the error also happens if CFLAGS=-fno-common is passed explicitly.

I know, but I'm not the one who cares about this being fixed to expend the effort. Maybe I phrased it rashly. What I meant was: to anyone invested in seeing this fixed, feel free to try it and let me know if it's still broken.

trofi:
Closing the report as invalid.

PR #77, opened 2020-02-15
Fix link OpenGL

Mizzrym1:
Sometimes OPENGL is found but OPENGL_gl_LIBRARY is not set because cmake is looking for two types OPENGL .

  Policy CMP0072 is not set: FindOpenGL prefers GLVND by default when
  available.  Run "cmake --help-policy CMP0072" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.

  FindOpenGL found both a legacy GL library:

    OPENGL_gl_LIBRARY: /usr/lib/x86_64-linux-gnu/libGL.so

  and GLVND libraries for OpenGL and GLX:

    OPENGL_opengl_LIBRARY: /usr/lib/x86_64-linux-gnu/libOpenGL.so
    OPENGL_glx_LIBRARY: /usr/lib/x86_64-linux-gnu/libGLX.so

  OpenGL_GL_PREFERENCE has not been set to "GLVND" or "LEGACY", so for
  compatibility with CMake 3.10 and below the legacy GL library will be used.

In my ubuntu 18.04 both work.


dcnieho:
Are you sure this is the right fix?
https://cmake.org/cmake/help/v3.10/module/FindOpenGL.html says "Clients are encouraged to use the OpenGL::* import targets instead." instead of the OPENGL_LIBRARIES variable

Mizzrym1:
Yes but in yours project "CMAKE_MINIMUM_REQUIRED(VERSION 3.0.0 FATAL_ERROR)"
OpenGL::* import targets are available from version 3.8.

rcmaniac25:
Good point, though according to https://cmake.org/cmake/help/v3.0/module/FindOpenGL.html, `OPENGL_gl_LIBRARY` and `OPENGL_glu_LIBRARY` have existed in every version along with `OPENGL_LIBRARIES`.

Perhaps it should instead be "leave as is, unless the CMake version is 3.8 and higher, then use OpenGL::* instead" as `OPENGL_LIBRARIES` seems to basically be deprecated or at least not recommended as of 3.10 and higher.

Mizzrym1:
> OPENGL_gl_LIBRARY and OPENGL_glu_LIBRARY have existed in every version along with OPENGL_LIBRARIES

Yes but https://cmake.org/cmake/help/v3.10/module/FindOpenGL.html says: 
If OPENGL_gl_LIBRARY is empty or not found and GLVND is available, the OpenGL::GL target will use GLVND OpenGL::OpenGL and OpenGL::GLX (and the OPENGL_LIBRARIES variable will use the corresponding libraries).

This means that OPENGL_LIBRARIES works even when OPENGL_gl_LIBRARY is not set.
I tested it on ubuntu 18.04 and it work on old and new CMake.
But if you want I can change this and check the CMake version and use OpenGL::GL if available.

jtsiomb:
If I'm reading the documentation correctly, `OPENGL_LIBRARIES` is not just the GL library, but also GLU, X11 and so on. Therefore, changing `OPENGL_gl_LIBRARY` to `OPENGL_LIBRARIES` would link GLU where it's not necessary, and will link X11 even if we do a wayland build.

As it stands, cmake produces a warning, but seems to work fine. Under which circuimstances did you find that `OPENGL_gl_LIBRARY` was not set correctly? I'd like to try and reproduce the original problem before changing anything.

Mizzrym1:
Example:
```
cat <<EOT > CMakeLists.txt
cmake_minimum_required(VERSION 3.20)
project(Test LANGUAGES C)

find_package(OpenGL)

message("OPENGL_FOUND: \${OPENGL_FOUND}")
message("OPENGL_gl_LIBRARY: \${OPENGL_gl_LIBRARY}")
message("OPENGL_opengl_LIBRARY: \${OPENGL_opengl_LIBRARY}")
message("OPENGL_glx_LIBRARY: \${OPENGL_glx_LIBRARY}")
message("OPENGL_LIBRARIES: \${OPENGL_LIBRARIES}")
message("OPENGL_INCLUDE_DIR: \${OPENGL_INCLUDE_DIR}")
EOT

docker run -it -v $(pwd):$(pwd) -w $(pwd) ubuntu bash -c "apt update && apt install -y cmake libopengl-dev libgl-dev && mkdir -p build && cd build && cmake .."
```
-- Found OpenGL: /usr/lib/x86_64-linux-gnu/libOpenGL.so   
OPENGL_FOUND: TRUE
OPENGL_gl_LIBRARY: 
OPENGL_opengl_LIBRARY: /usr/lib/x86_64-linux-gnu/libOpenGL.so
OPENGL_glx_LIBRARY: /usr/lib/x86_64-linux-gnu/libGLX.so
OPENGL_LIBRARIES: /usr/lib/x86_64-linux-gnu/libOpenGL.so;/usr/lib/x86_64-linux-gnu/libGLX.so
OPENGL_INCLUDE_DIR: /usr/include

Open gl is found but OPENGL_gl_LIBRARY is empty

jtsiomb:
I tried your cmake file and indeed OPENGL_gl_LIBRARY is undefined, but if I change the version to 3.0.0, as it is in the freeglut cmake file,  then it works and points to libGL. If I keep the version to 3.20, but also set OpenGL_GL_PREFERENCE to LEGACY, it also works and points to libGL.

So as long as we don't change the required version, it seems like this is never going to be an issue, and if we do at some point up the required version, we can just set that option and we'll get the same result, making this a total non-issue.

The real issue here seems to be something else. It's ok for an application to just ask for "any opengl" and get one or the other by using OPENG_LIBRARIES as you proposed, but what happens if freeglut links with libGL and the application with libOpenGL, or vice-versa? Should we make this a build option and fork yet another installable freeglut variant? Should we stop linking with OpenGL and resolve symbols at runtime from whichever library the application linked? Should THAT be a build-time option too? I'm not in the mood for tackling these deeper problems at the moment, and this pull request is not about that anyway, so for now I'll just close this, and leave that whole mess for another time.

Mizzrym1:
Ok but i don't understand why freeglut must link OPENGL_gl_LIBRARY when it work with both options.
I use freglut as subdirectory and this is problem because i require newer cmake version and i need find OpenGL too.
I spend many time for find the problem. And I wanted fix this problem for other.

jtsiomb:
Most people are using freeglut and GLUT before that, by just adding `-lGL -lglut` to their linker command line. We cannot break that usage. Not everyone is using cmake and submodules. If freeglut suddenly started linking with `libOpenGL` and conflict with that usage, that would be unacceptable. A sub-optimal solution which does not break current use patterns is preferable, and if we have to pick one of "link with libGL" or "link with libOpenGL", we should link with libGL.

As I said, ideally at some point we'll have to find a way to either be compatible with all possible use cases, or at least make it a build option. But that's a story for another time.

PR #80, opened 2020-02-17
Fix LIBNAME on MSVC

Mizzrym1:
On MSVC static lib and import from shared lib have the same name


jtsiomb:
I guess this slipped by since I usually don't touch windows stuff, but noone else responded either. This looks reasonable so I'll just go ahead and merge it.

PR #81, opened 2020-02-17
Target include directories

Mizzrym1:
Add include directories to target for add FreeGLUT like "add_subdirectory".
Move configured file from CMAKE_BINARY_DIR to CMAKE_CURRENT_BINARY_DIR.

### DIFF
--- a/freeglut/freeglut/CMakeLists.txt
+++ b/freeglut/freeglut/CMakeLists.txt
@@ -135,13 +135,13 @@ IF(WIN32)
         src/mswin/fg_state_mswin.c
         src/mswin/fg_structure_mswin.c
         src/mswin/fg_window_mswin.c
-        ${CMAKE_BINARY_DIR}/freeglut.rc # generated below from freeglut.rc.in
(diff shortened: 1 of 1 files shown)


dcnieho:
Thanks for the contribution. What problem is this fixing?

Mizzrym1:
When I add freeglut as subdirectory in my project and freeglut target has set TARGET_INCLUDE_DIRECTORIES i can only link to this target. I don't must look for what include directories I need add.

When I add freeglut as subdirectory in my project CMAKE_BINARY_DIR is binary directory from my project. CMAKE_CURRENT_BINARY_DIR is binary directory for freeglut project.
When you have CMAKE_BINARY_DIR instead CMAKE_CURRENT_BINARY_DIR you overwrite my file config.h.

stefanbuettner:
I ran into the same problem that the freeglut[_static] targets do not have the INTERFACE_INCLUDE_DIRECTORIES set properly. This change fixed it for me.

I was surprised that this change's history is unrelated to the v3.4.0 tag as I was trying to merge the latest release in.

Eventually I recreated the fix on v3.4.0 and used that modified version: https://github.com/stefanbuettner/freeglut/tree/add-target-include-directories

PR #82, opened 2020-02-17
Remove default FREEGLUT_LIB_PRAGMAS

Mizzrym1:
Why are you linking the library from the header file?
I lost two hours not understanding why my program wants to link a library with a different name than it was built.
Can you delete this option by default?

### DIFF
--- a/freeglut/freeglut/include/GL/freeglut_std.h
+++ b/freeglut/freeglut/include/GL/freeglut_std.h
@@ -47,11 +47,7 @@
  * The default behavior depends on the compiler/platform.
  */
 #   ifndef FREEGLUT_LIB_PRAGMAS
-#       if ( defined(_MSC_VER) || defined(__WATCOMC__) ) && !defined(_WIN32_WCE)
-#           define FREEGLUT_LIB_PRAGMAS 1
-#       else
-#           define FREEGLUT_LIB_PRAGMAS 0
-#       endif
+#       define FREEGLUT_LIB_PRAGMAS 0
 #   endif
 
 #  ifndef WIN32_LEAN_AND_MEAN


rcmaniac25:
So, the problem with this is the primary goal of the freeglut project is to maintain compatibility with GLUT.

On Windows, glut (3.7) goes the following:
```
#pragma comment (lib, "winmm.lib")     /* link with Windows MultiMedia lib */
#pragma comment (lib, "opengl32.lib")  /* link with Microsoft OpenGL lib */
#pragma comment (lib, "glu32.lib")     /* link with OpenGL Utility lib */
#pragma comment (lib, "glut32.lib")    /* link with Win32 GLUT lib */
```

So we can't remove this to the best of my knowledge.

If you really think it should be removed and wouldn't cause compatibility issues, I'd suggest asking on freeglut-developer mailing list since this is a mirror of the repo and we don't tend to respond super quick here.

rcmaniac25:
Just noticed I had a slightly newer glut (3.7.6):
```
/* To disable automatic library usage for GLUT, define GLUT_NO_LIB_PRAGMA
   in your compile preprocessor options. */
# if !defined(GLUT_BUILDING_LIB) && !defined(GLUT_NO_LIB_PRAGMA)
#  pragma comment (lib, "winmm.lib")      /* link with Windows MultiMedia lib */
/* To enable automatic SGI OpenGL for Windows library usage for GLUT,
   define GLUT_USE_SGI_OPENGL in your compile preprocessor options.  */
#  ifdef GLUT_USE_SGI_OPENGL
#   pragma comment (lib, "opengl.lib")    /* link with SGI OpenGL for Windows lib */
#   pragma comment (lib, "glu.lib")       /* link with SGI OpenGL Utility lib */
#   pragma comment (lib, "glut.lib")      /* link with Win32 GLUT for SGI OpenGL lib */
#  else
#   pragma comment (lib, "opengl32.lib")  /* link with Microsoft OpenGL lib */
#   pragma comment (lib, "glu32.lib")     /* link with Microsoft OpenGL Utility lib */
#   pragma comment (lib, "glut32.lib")    /* link with Win32 GLUT lib */
#  endif
# endif
```

Mizzrym1:
Compatibility. :(
I understand.
But when I build my project in debug it doesn't mean that freeglut is also a debug.
I would like to be able to find GLUT in cmake and simply link to the GLUT::GLUT target.
Don't wonder what definitions I need to set.

rcmaniac25:
Usually libraries come with a debug and release lib for windows. If not, then debug/release point to the same library as a find function or from IDE search paths.

If you'd like to change it, as stated before, you'll have to bring it up in the freeglut-developer mailing list.

Mizzrym1:
> Usually libraries come with a debug and release lib for windows.

Ok, but not freeglut. freeglut has only release. And probably manually edited file "freeglut_std.h". 
Anyone who wants to build must change it manually?

> So, the problem with this is the primary goal of the freeglut project is to maintain compatibility with GLUT.

It is for compatibility with GLUT, but it don't work witch option "FREEGLUT_REPLACE_GLUT" because has wrong name of library. And without this option I don't have a file "include/GL/glut.h" :)

jtsiomb:
Even if compatibility was no factor, I don't see why this change would be necessary. Why not just define FREEGLUT_LIB_PRAGMAS to 0 if you don't want the pragma linker directives?

PR #83, opened 2020-05-13
Change GLUT_CURSOR_INFO from IDC_HELP to IDC_HAND

Schneegans:
This is a very simple fix. According to https://www.opengl.org/resources/libraries/glut/spec3/node28.html, `GLUT_CURSOR_INFO` should be a "Pointing Hand". `IDC_HELP` is however an arrow with a question mark; `IDC_HAND` gives the cursor image I am expecting.

Is there a reason for  `GLUT_CURSOR_INFO` being set to `IDC_HELP` (wich is also correctly used for ` GLUT_CURSOR_HELP`)?


jtsiomb:
Makes sense ...
Merged into master branch.

PR #88, opened 2020-12-10
spaceball_support_for_spacemouse_bug_258

saxondigital:
Supports the new 3Dconnexion VID=0x256f devices and fixes bugs for the existing support for Logitech VID = 0x046d (legacy 3Dconnexion devices). Tested on Space Navigator, Space Mouse Pro (original), SpaceMouse Wireless and SpaceMouse Enterprise.

### DIFF
--- a/freeglut/freeglut/src/mswin/fg_main_mswin.c
+++ b/freeglut/freeglut/src/mswin/fg_main_mswin.c
@@ -1544,13 +1544,14 @@ LRESULT CALLBACK fgPlatformWindowProc( HWND hWnd, UINT uMsg, WPARAM wParam, LPAR
     }
 #endif
 
-#ifdef WM_INPUT
+#ifdef WM_INPUT	
+	case WM_INPUT_DEVICE_CHANGE:	/* Added by Shane Saxon. Dec 2020 */
+		fgSpaceballDeviceChangeWinEvent( hWnd, wParam, lParam );
+        break;
     case WM_INPUT:
         /* Added by Jinrong Xie <stonexjr at gmail.com> for SpaceNavigator support on Windows. Dec 2014 */
-        if (fgHasSpaceball())
-        {
-            fgSpaceballHandleWinEvent(hWnd, wParam, lParam);
-        }
+        if( fgHasSpaceball() )
+			fgSpaceballHandleWinEvent( hWnd, wParam, lParam );
         break;
 #endif
     default:
--- a/freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
+++ b/freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
@@ -11,6 +11,10 @@
  * Written by Jinrong Xie <stonexjr at gmail.com>
  * Modification date: Wed Dec 24, 2014
  *
+ * Copyright (c) 2020 Shane Saxon. All Rights Reserved.
+ * Written by Shane Saxon <saxon at openantz.com>
+ * Modification date: Fri Dec 4, 2020
+ *
  * Permission is hereby granted, free of charge, to any person obtaining a
  * copy of this software and associated documentation files (the "Software"),
  * to deal in the Software without restriction, including without limitation
@@ -34,6 +38,10 @@
  * for Space Navigator support on Windows.
  * This code is enhanced by at least supporting 3Dconnexion's
  * six degree of freedom navigator.
+ *
+ * Modified by Shane Saxon 2020-12-07, to support (new) VID=0x256f devices
+ * and enhanced original VID=0x046d support. Tested on Space Navigator,
+ * Space Mouse Pro (original), SpaceMouse Wireless and SpaceMouse Enterprise.
  */
 
 #if(_WIN32_WINNT >= 0x0501)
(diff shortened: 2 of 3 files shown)


freeglut/freeglut/src/mswin/fg_main_mswin.c
```diff
@@ -1544,13 +1544,14 @@ LRESULT CALLBACK fgPlatformWindowProc( HWND hWnd, UINT uMsg, WPARAM wParam, LPAR
     }
 #endif
 
-#ifdef WM_INPUT
+#ifdef WM_INPUT
```
jtsiomb, line 1547: whitespace at the end of line

freeglut/freeglut/src/mswin/fg_main_mswin.c
```diff
@@ -1544,13 +1544,14 @@ LRESULT CALLBACK fgPlatformWindowProc( HWND hWnd, UINT uMsg, WPARAM wParam, LPAR
     }
 #endif
 
-#ifdef WM_INPUT
+#ifdef WM_INPUT	
+	case WM_INPUT_DEVICE_CHANGE:	/* Added by Shane Saxon. Dec 2020 */
+		fgSpaceballDeviceChangeWinEvent( hWnd, wParam, lParam );
```
jtsiomb, line 1549: Match the indentation of the surrounding block. The rest of the code in this part uses spaces, not tabs.

freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
```diff
@@ -1,199 +1,612 @@
-/*
```
jtsiomb, line 1: This whole file used to have UNIX line endings, and your patch changes it to DOS line endings, making it extremely hard to see what the actual changes are. The whole file is treated as completely replaced by git as you can see. Please fix the line endings to make it possible to pinpoint changes, and to review it properly.

jtsiomb (changes requested):

saxondigital:
All tabs replaced with spaces and CRLF line endings conveted to LF, (I used the 'dos2unix' utility to convert line endings, which is convienantly available on all platforms). There should (now) not be any whitespace at the end of the lines, (was likely the CRLF).

`fg_spaceball_mswin.c` is nearly completely re-written. In addition to new device support, it has more robust error handling and implements previously empty functions. All (exposed) `// C++ (or C99) style comments` changed to `/* C style comments */` for wider compiler support. Also has some commented out bits that could be useful to others, such as the key codes listed at the end of the file.

Overall, the style is similar to what was originally in the file (and gleamed from other code in the project). A bit of guess work exists though, (naming conventions, etc) and button handling (key code values are passed through directly, regardless of event type). Eg. was wrestling with naming the event (report ID) constants, and chose them as such to make it clear they are device specific. But on the flip-side, was thinking about ways to make it more congruent with the X11 counter-part.

freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
```diff
@@ -42,158 +50,563 @@
...
+         */
+        uint32_t prevBDF = 0;
+        uint32_t nowBDF = *(uint32_t*)(&pRawInput->data.hid.bRawData[1]);
+
+        if( FETCH_WCB(*window, SpaceButton) )
+        {
+            int i;
+            for( i=0; i < 32; i++ )
+            {
+                bool stateBefore = (prevBDF >> i) & 1;
```
jtsiomb, line 430: C89 doesn't have bool

freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
```diff
@@ -42,158 +50,563 @@
...
-    HWND hwnd;
-    fg_sball_initialized = 1;
-    if (!fgStructure.CurrentWindow)
+    if( fg_sball_initialized == 1 )
+        return;
+
+    if( !fgStructure.CurrentWindow )
     {
-        fg_sball_initialized = 0;
+        fg_sball_initialized = -1;
```
jtsiomb, line 124: Here you changed the initialization of fg_sball_initialized to -1, but further on all tests are `if(!fg_sball_initialized)`, am I missing something?

freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
```diff
@@ -42,158 +50,563 @@
...
-        {
-            return;
-        }
+        if( !fg_sball_initialized )
+             return;
     }
 
-    res = GetRawInputData(hRawInput, RID_INPUT, NULL, &size, sizeof(RAWINPUTHEADER));
-    if (res == -1)
+    if( GetRawInputData( hRawInput, RID_INPUT, NULL, &size,
```
jtsiomb, line 295: random formatting change

freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
```diff
@@ -42,158 +50,563 @@
...
+            sizeof(RAWINPUTHEADER) ) == -1 )
         return;
 
-    rawInputBuffer = malloc(size * sizeof *rawInputBuffer);
```
jtsiomb, line 129: Why change `sizeof *rawInputBuffer` to `sizeof(void*)` ?

freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
```diff
@@ -42,158 +50,563 @@
...
-    UINT size;
-    BYTE *rawInputBuffer;
     PRAWINPUT pRawInput;
-    UINT res;
     RID_DEVICE_INFO sRidDeviceInfo;
+    BYTE *rawInputBuffer;
+    UINT size;
 
-    if (!fg_sball_initialized)
+    if( !fg_sball_initialized )
```
jtsiomb, line 288: random formatting change

freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
```diff
@@ -42,158 +50,563 @@
...
     {
         fgPlatformInitializeSpaceball();
-        if (!fg_sball_initialized)
-        {
-            return;
-        }
+        if( !fg_sball_initialized )
```
jtsiomb, line 291: random formatting change

freeglut/freeglut/src/mswin/fg_main_mswin.c
```diff
@@ -1545,12 +1545,13 @@ LRESULT CALLBACK fgPlatformWindowProc( HWND hWnd, UINT uMsg, WPARAM wParam, LPAR
...
+    case WM_INPUT_DEVICE_CHANGE: /* Added by Shane Saxon. Dec 2020 */
+        fgSpaceballDeviceChangeWinEvent( hWnd, wParam, lParam );
+        break;
     case WM_INPUT:
         /* Added by Jinrong Xie <stonexjr at gmail.com> for SpaceNavigator support on Windows. Dec 2014 */
-        if (fgHasSpaceball())
-        {
-            fgSpaceballHandleWinEvent(hWnd, wParam, lParam);
-        }
+        if( fgHasSpaceball() )
```
jtsiomb, line 1553: random formatting change

jtsiomb (changes requested):
Now this is much better. With just the changes highlighted, I managed to take a closer look. I've noticed two potential mistakes, and a number of random formatting changes with no reason. I've marked them with review comments.

Since a few days passed and nobody seems to be raising any objections on the mailing list, as soon as these minor things are addressed, I'm fine with merging these changes.

saxondigital:
I changed `fg_sball_initialized = -1;` based on some of the X11 code.  But, I think that reverting back to just using 0 makes more sense.

Regarding the sizeof(void*), this is just my default style, I will double check this malloc and likely revert it to using sizeof(*rawInputBuffer).

The rest of your comments look good as well, (most of the random format changes are a result of cleaning up the entire file).  I plan to integrate your feedback, commit the changes and will generate a new pull request shortly.

freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
saxondigital: okay, both are wrong, I checked the docs for GetRawInputData() and the pcbSize parameter is: `The size, in bytes, of the data in pData.`
So will make it: `malloc( size * sizeof(BYTE) );`

freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
saxondigital: This was changed to maintain the code style/formatting througout the file. I prefer to keep things consistent.  But let me know if you would still like to revert it.

freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
saxondigital: nope, you caught something! Changing back to = 0.

freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
saxondigital: hmm, this was changed to keep the style consistent with the function`fgSpaceballUpdateDeviceList()`. If you like, I can revert it, but prefer to then do the same to all 6 of the `if(... == -1) return;` statements.
verdict?

freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
saxondigital: good point, will use plain ol' type int.

freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
saxondigital: will revert to the original verbose curly brackets.

freeglut/freeglut/src/mswin/fg_main_mswin.c
saxondigital: will revert to original formatting.

freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
saxondigital: I wanted to add that this is not just formatting, and was part of the compiled binary size (memory) optimizations.  This one saved about 100 bytes, and was part of a larger effort (that brought the binary size down to 222KB from 224KB). Note that the DLL is currently 1KB larger than before I started, which seems pretty good considering the addtional functionality.

jtsiomb: I don't understand. How can a formatting change affect the filesize?

Also my comments about formatting changes have nothing to do with what I prefer, or even about consistency. I just don't want to see changes in lines where you didn't change anything functional, because that muddles up the change history and makes it harder to track what each commit affected.

freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
jtsiomb: sizeof(BYTE) is by definition 1, so leave it out.

saxondigital: indeed, sounds good = efficient.

freeglut/freeglut/src/mswin/fg_spaceball_mswin.c
saxondigital: checking the result of the (MSW) HID functions occurs in two functions, each requiring allocating the 'res' variable (happens in 2 functions = 2 extra allocations), copying it from the return value and then checking it (6 occurences).  I have not directly compared the compiled machine code, but in general have found that this micro-optimization does save a little (all together I guesstimate about 100 bytes or less).  Also, this was how the HID sample code I drew upon was handling the err checking.  But in the end, it makes little difference to me (and only a tiny difference to the binary size.)
your call...

saxondigital: okay, i think were actually talking about two different lines of code... I believe I have applied the requested changes (plus similar).

saxondigital:
okay, I went through the whole file looking for 'random formatting change' occurences and found a few others (and reverted them too).  Also, applied a minor optimization (and the .dll compiled 1/2 KB smaller).

let me know what you think!

Issue #70, opened 2019-01-26
failed to compile freeglut on macOS 10.14.2


VVingerfly:
Thank you for submitting an issue, it's much appreciated.

Couple points before you create this ticket:

* Because this is a mirror repo, we may not respond to new issues in the timely manner you may expect.
* Our official Bug/Issue list is located [here](https://sourceforge.net/p/freeglut/bugs/). If you create a duplicate issue in the official list in addition to GitHub, please reference the GitHub and SourceForge tickets in the other site's ticket so we may close both tickets.
* There is also an official Feature Request list, located [here](https://sourceforge.net/p/freeglut/feature-requests/).
* One way to get more attention to your issue is to message our mailing list: `freeglut-developer at lists.sourceforge.net`

If you still intend to make a ticket here, please provide the following:

* OS: macOS 10.14.2
* FreeGLUT version: git_master branch

Hi, I want to compile freeglut on my MBP. But it failed with the following message

```
WMBP:build lw$ cmake ..
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/lw/Desktop/freeglut_github/freeglut/freeglut/build
WMBP:build lw$ make
[ 30%] Built target freeglut_static
[ 31%] Linking C executable bin/timer_callback_static
Undefined symbols for architecture x86_64:
  "_glXChooseFBConfig", referenced from:
      _fghChooseConfig in libglut.a(fg_window_x11_glx.c.o)
      _fgPlatformGlutGetModeValues in libglut.a(fg_state_x11_glx.c.o)
  "_glXCreateNewContext", referenced from:
      _fghCreateNewContext in libglut.a(fg_window_x11_glx.c.o)
  "_glXDestroyContext", referenced from:
      _fgPlatformDestroyContext in libglut.a(fg_init_x11.c.o)
      _fgPlatformCloseWindow in libglut.a(fg_window_x11.c.o)
  "_glXGetCurrentContext", referenced from:
      _fgPlatformOpenWindow in libglut.a(fg_window_x11.c.o)
  "_glXGetFBConfigAttrib", referenced from:
      _fghChooseConfig in libglut.a(fg_window_x11_glx.c.o)
      _fgPlatformGetConfig in libglut.a(fg_state_x11_glx.c.o)
      _fgPlatformGlutGetModeValues in libglut.a(fg_state_x11_glx.c.o)
  "_glXGetProcAddressARB", referenced from:
[...]

PR #89, opened 2021-02-19
Find correct X OpenGL libraries on Mac

multiplemonomials:
I ran into some issues building FreeGLUT on Mac.  Looks like the issue is really with CMake not providing a way to get the X11 OpenGL libraries, but I figured out a workaround that removes the need for manually editing cache variables.  

First of all, I moved the block that finds X11 up before OpenGL.  Then, when it detects a mac build, it finds the path where the X11 libraries are and manually checks for libGL.dylib and libGLU.dylib at this location.  If they're found, they're saved into the correct cache variables so that FindOpenGL will use them.

This partially fixes GitHub #70 and [SourceForge bug 218](https://sourceforge.net/p/freeglut/bugs/218/).  On OS X 10.9 (the only version I can test), it is now able to build shared FreeGLUT and many of the examples without any issues or manual changes needed.  (However, there are still some link errors with the static version that I didn't look into).

### DIFF
--- a/freeglut/freeglut/CMakeLists.txt
+++ b/freeglut/freeglut/CMakeLists.txt
@@ -254,6 +254,28 @@ IF(FREEGLUT_GLES OR FREEGLUT_WAYLAND)
     )
 ENDIF()
 
+INCLUDE(CheckIncludeFiles)
+IF(UNIX AND NOT(ANDROID OR BLACKBERRY OR FREEGLUT_WAYLAND))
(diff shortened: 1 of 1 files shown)


multiplemonomials:
@dcnieho Have you had a chance to take a look at this?

jtsiomb:
Thank you, I just merged this to the master branch.

This repository used to be a mirror, and contributions here were not guaranteed to be noticed. Since it has become the official one, I'm *very slowly* going through old issues and pull requests.

PR #91, opened 2021-09-10
Fix illegal instruction in glutTimerFuncUcall()

y-guyon:
When fgState.Timers.First is NULL, it goes straight to the fgListInsert() call
and tries to dereference &node->Node, with node being NULL, which is illegal.

I do not know how it was working before because searching for "Timers" on
the whole repository does not indicate any other place where fgState.Timers
is assigned anything but NULL, so the first call to glutTimerFuncUcall() should
have been an issue. Maybe compilers or runtime checks changed recently.
Did I miss something?


jtsiomb:
merged manually on the new master branch

PR #95, opened 2022-01-16
Change define WIN32_LEAN_AND_MEAN for windows.h in GCC

Jamaika1:


jtsiomb:
What problem does this fix?

jtsiomb:
This doesn't seem like a useful change, it doesn't come with any information about what it's supposed to fix, and the OP is not responding. Closing.

PR #98, opened 2022-03-29
mingw_cross: Add C++ compiler and remove deprecated CMakeForceCompiler

tormodvolden:
I must say I don't know what the `Prefix detection only works with compiler id "GNU"` was about, but it was there since 2012, and it seems not needed any longer.

I have tested the resulting libglut.dll on Windows 10 (with libfreenect-glview).

### DIFF
--- a/mingw_cross_toolchain.cmake
+++ b/mingw_cross_toolchain.cmake
@@ -1,9 +1,7 @@
 # CMake toolchain file, cf. README.mingw_cross
 SET(CMAKE_SYSTEM_NAME Windows)
-include(CMakeForceCompiler)
 IF("${GNU_HOST}" STREQUAL "")
-	SET(GNU_HOST i586-mingw32msvc)
+	SET(GNU_HOST x86_64-w64-mingw32)
 ENDIF()
-# Prefix detection only works with compiler id "GNU"
-CMAKE_FORCE_C_COMPILER(${GNU_HOST}-gcc GNU)
+SET(CMAKE_C_COMPILER  ${GNU_HOST}-gcc)
 SET(CMAKE_RC_COMPILER ${GNU_HOST}-windres)


jtsiomb (changes requested):
I agree with the change of `GNU_HOST` to a more reasonable default. I also tested it, and even with cmake 3.0.0, which is our earliest supported cmake version, forcing the compiler doesn't seem to be necessary. So I'm ok with merging this to remove the warning. But there are a couple of changes I want first:

  1. Adding `CMAKE_CXX_COMPILER` is not the Right Thing to do. The root problem was that we had a single demo with an erroneous ".cpp" suffix, which I fixed just now. So this shouldn't be needed any more.

  2. Since we're not using `CMAKE_FORCE_C_COMPILER` any more, the `include(CMakeForceCompiler)` line needs to go.

  3. Finally, why did you swap the order of gcc and windres? Not an issue, but that's an pretty arbitrary change, so let's avoid it ideally.

tormodvolden:
Hi, thanks for the review! I didn't expect this so fast, so I already pushed a fixup before seeing your comment.

1. OK, sounds good, let's get rid of that then.
2. Was already taken away in my repush
3. No reason other than esthetics. Believe me, the order of these two does not affect anything.

tormodvolden:
Heh, it was not for esthetics. Now everything should be good.

mingw_cross_toolchain.cmake
```diff
@@ -1,9 +1,10 @@
 # CMake toolchain file, cf. README.mingw_cross
+
```
dcnieho, line 2: why are there a bunch of new empty lines? could you clean that up?

tormodvolden: Readability! Looks more like other scripts I have seen. But I can adapt to your liking.

tormodvolden: done

tormodvolden:
Thanks for merging it. BTW I think it is possible to configure GitHub to do fast-forward merging to avoid meaningless merge commits, if you would prefer that.

jtsiomb:
I'm not too fussed to be honest, either way it's clear what's happening. But it might be slightly cleaner, so I'll look into it next time.

PR #103, opened 2022-05-19
make FREEGLUT_BUILD_DEMO OFF

katahiromz:
Reduce the compilation time for many regular people.


jtsiomb:
This feels like a completely arbitrary change. It is an option, anyone who doesn't want to build the demos can turn the option off. I don't see the point in changing the default.

PR #105, opened 2022-05-20
Fix 'suggest explicit braces' warning

katahiromz:
On MSYS2 gcc:
```txt
C:/Users/katahiromz/Desktop/freeglut/src/mswin/fg_main_mswin.c: In function 'fghWindowProcKeyPress':
C:/Users/katahiromz/Desktop/freeglut/src/mswin/fg_main_mswin.c:723:7: warning: suggest explicit braces to avoid ambiguous 'else' [-Wdangling-else]
  723 |     if( keypress != -1 )
      |       ^
```


jtsiomb (changes requested):
That's way too much. I agree that the outer set of braces would make it slightly clearer, but the inner ones aren't necessary.

PR #111, opened 2022-07-26
Fix aliasing violation

mattst88:
Noticed when compiling with link-time optimizations.

````
include/GL/freeglut_std.h:240:18: error: type of ‘glutBitmapHelvetica18’ does not match original declaration [-Werror=lto-type-mismatch]
  240 |     extern void* glutBitmapHelvetica18;
      |                  ^
src/x11/fg_glutfont_definitions_x11.c:103:27: note: ‘glutBitmapHelvetica18’ was previously declared here
  103 | struct freeglutBitmapFont glutBitmapHelvetica18 ;
      |                           ^
src/x11/fg_glutfont_definitions_x11.c:103:27: note: code may be misoptimized unless ‘-fno-strict-aliasing’ is used
````

### DIFF
--- a/include/GL/freeglut_std.h
+++ b/include/GL/freeglut_std.h
@@ -227,15 +227,18 @@
     /*
      * I don't really know if it's a good idea... But here it goes:
      */
-    extern void* glutStrokeRoman;
-    extern void* glutStrokeMonoRoman;
-    extern void* glutBitmap9By15;
-    extern void* glutBitmap8By13;
-    extern void* glutBitmapTimesRoman10;
-    extern void* glutBitmapTimesRoman24;
-    extern void* glutBitmapHelvetica10;
-    extern void* glutBitmapHelvetica12;
-    extern void* glutBitmapHelvetica18;
+
+    struct freeglutStrokeFont;
+    struct freeglutBitmapFont;
+    extern struct freeglutStrokeFont glutStrokeRoman;
+    extern struct freeglutStrokeFont glutStrokeMonoRoman;
+    extern struct freeglutBitmapFont glutBitmap9By15;
+    extern struct freeglutBitmapFont glutBitmap8By13;
+    extern struct freeglutBitmapFont glutBitmapTimesRoman10;
+    extern struct freeglutBitmapFont glutBitmapTimesRoman24;
+    extern struct freeglutBitmapFont glutBitmapHelvetica10;
+    extern struct freeglutBitmapFont glutBitmapHelvetica12;
+    extern struct freeglutBitmapFont glutBitmapHelvetica18;
 
     /*
      * Those pointers will be used by following definitions:


jtsiomb:
That's part of the public API of freeglut. It's set in stone. Even more so, it's part of the original GLUT public API, where the fonts are also defined as `void*`, so it's doubly set in stone.

If there's something that should change to fix this, it should be in the implementation, not the public header files.

mattst88:
I'm not super familiar with the GLUT API, so maybe you can help me understand a bit better.

I see that e.g. `glutStrokeRoman` is used only in the definition of `GLUT_STROKE_ROMAN`. I agree that `GLUT_STROKE_ROMAN` is part of the ABI, but it is actually not changing since it is explicitly cast to `void *`—-`((void *) &glutStrokeRoman)`

Why do you say that `glutStrokeRoman` itself is part of the public API? How can it be used outside of `GLUT_STROKE_ROMAN`?

jtsiomb:
Certainly, let me elaborate on my previous comment:

Your pull request changes a public header file: `freeglut_std.h`. In that header file there is a declaration of a pointer `glutStrokeRoman`. Your change makes it so that `glutStrokeRoman` is no longer a pointer. This introduces incompatibilities with any existing code which expects it to be a `void*`.

This does not only affect binary compatibility with previous releases of freeglut, but binary compatibility with the original GLUT library too, which also defines these symbols as `void*`.

We do not break backwards compatibility, either with previous releases of freeglut, or with the original GLUT, at least not on purpose. Therefore, in perpetuity, the symbol `glutStrokeRoman` will remain in the freeglut public API, declared as `void*`, until the sun explodes and incinerates every last bit of the freeglut source code.

If the existing situation causes an issue (which btw you only theorized about, and have not demonstrated), there are two possible solutions:
   1. live with it.
   2. change the internal implementation of freeglut.

jtsiomb:
Also, I don't even understand how your change is supposed to compile. You add a forward declaration for an opaque structure, and then you attempt to declare objects of that structure, with unknown to the compiler size. Did you try your change before issuing a pull request?

mattst88:
> In that header file there is a declaration of a pointer glutStrokeRoman. Your change makes it so that glutStrokeRoman is no longer a pointer. This introduces incompatibilities with any existing code which expects it to be a void*.

My question is whether the symbol `glutStrokeRoman` is actually part of the API (or ABI).

I would agree that it is if, for example, a function took it as an argument, but as far as I'm aware that's not the case—functions only take `GLUT_STROKE_ROMAN`. We can be sure of this because `glutStrokeRoman` isn't provided on all platforms.

My claim is that no code uses `glutStrokeRoman` directly, so it's not part of the API. I'd have to do some thinking about whether its part of the ABI, but I kind of don't think it's part of the ABI either since you just need the symbol's address.

mattst88:
> Also, I don't even understand how your change is supposed to compile. You add a forward declaration for an opaque structure, and then you attempt to declare objects of that structure, with unknown to the compiler size. Did you try your change before issuing a pull request?

Yes, it builds fine.

I believe it's legal because we're not actually declaring an instance of the `struct` in the header since it's marked `extern`. Then we just take the address of it, which is fine.

jtsiomb:
If it is in a public header, it's part of the public API.

"Other platforms" have them too, because GLUT 3.7 has them. The de-facto standard for binary compatibility with GLUT, is to be compatible with the original GLUT implementation. There's no official standard.

mattst88:
> If it is in a public header, it's part of the public API.

I get what you're saying, but if a tree falls in the woods...

jtsiomb:
This is not up for discussion. I'm just explaining the facts.

It does not matter if I can conceptualize a use case for using that pointer directly. It's there declared the way it is, for binary compatibility with any perverse use a program might have for it from the times of the original GLUT implementation.

PR #112, opened 2022-07-26
Fix aliasing violation

mattst88:
Noticed when compiling with link-time optimizations.

````
include/GL/freeglut_std.h:240:18: error: type of ‘glutBitmapHelvetica18’ does not match original declaration [-Werror=lto-type-mismatch]
  240 |     extern void* glutBitmapHelvetica18;
      |                  ^
src/x11/fg_glutfont_definitions_x11.c:103:27: note: ‘glutBitmapHelvetica18’ was previously declared here
  103 | struct freeglutBitmapFont glutBitmapHelvetica18 ;
      |                           ^
src/x11/fg_glutfont_definitions_x11.c:103:27: note: code may be misoptimized unless ‘-fno-strict-aliasing’ is used
````

### DIFF
--- a/src/x11/fg_glutfont_definitions_x11.c
+++ b/src/x11/fg_glutfont_definitions_x11.c
@@ -91,14 +91,25 @@ struct freeglutBitmapFont
 };
 
 
-struct freeglutStrokeFont glutStrokeRoman ;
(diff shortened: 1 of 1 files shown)


jtsiomb:
much better

Issue #113, opened 2022-08-04
Install Error: file INSTALL cannot find `freeglutd.pdb` with `Ninja Multi-Config` generator.


hwhsu1231:
### Descriptions

I tried to compile the source code. It turns out that errors occurred in **INSTALL** stages with `freeglutd.pdb`. However, there indeed exists `freeglutd.pdb` inside the build folder.

```cmd
D:\.ccxxpkgs\source\freeglut-3.2.2\build-msvc16-x64>cmake --install . --config Debug
-- Installing: D:/.ccxxpkgs/install/msvc16-x64/freeglut-3.2.2/lib/freeglutd.lib
-- Installing: D:/.ccxxpkgs/install/msvc16-x64/freeglut-3.2.2/bin/freeglutd.dll
CMake Error at cmake_install.cmake:57 (file):
  file INSTALL cannot find
  "D:/.ccxxpkgs/source/freeglut-3.2.2/build-msvc16-x64/bin/freeglutd.pdb":
  File exists.
```

### Steps of Installation

<details><summary>Click to expand</summary>

1.  Download and unzip the source code.

    [freeglut-3.2.2.zip](https://github.com/FreeGLUTProject/freeglut/archive/refs/tags/v.zip)

2.  Open Windows Terminal (`cmd.exe`) and change to the project folder.

    ```
    cd freeglut-3.2.2
    ```

3.  call `vcvarsall.bat x64` to preload the MSVC environment.

    ```
    vcvarsall x64
    ```

4.  Create the build folder and change into it.

    ```
    md build-msvc16-x64
    cd build-msvc16-x64
    ```

5.  Configure the project with the generator `Ninja Multi-Config`.

    ```
    cmake .. -G"Ninja Multi-Config"
    ```

6.  Open `cmake-gui` to edit some Cache Variables. After editing, press **Configure** and **Generate** buttons. Then, close it.

    ```
    cmake-gui .

    CMAKE_INSTALL_PREFIX=D:/.ccxxpkgs/install/msvc16-x64/freeglut-3.2.2
    ```

7.  Build the project with `Debug` configuration.

    ```
    cmake --build . --config Debug
    ```

8.  Install the project with `Debug` configuration.

    ```
    cmake --install . --config Debug
    ```

</details>

### Logs

<details><summary>Click to expand</summary>

```cmd
D:\.ccxxpkgs\source\freeglut-3.2.2>md build-msvc16-x64

[...]

PR #114, opened 2022-08-04
Support other Multi-Config generators with Debug.

hwhsu1231:
Try to fix: #113

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -509,7 +509,8 @@ IF(FREEGLUT_BUILD_SHARED_LIBS)
             INCLUDES DESTINATION include
     )
     IF(INSTALL_PDB)
-        IF(CMAKE_GENERATOR MATCHES "^Visual Studio")
+        get_cmake_property(is_multi_config GENERATOR_IS_MULTI_CONFIG)
+        IF(is_multi_config)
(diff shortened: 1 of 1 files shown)


jtsiomb:
Interesting. `GENERATOR_IS_MULTI_CONFIG` is only available from cmake 3.9. If I'm reading the cmake docs correctly, older versions of cmake which don't have that property will just return `NOTFOUND` which I'm not sure what it does when fed to that IF. We should handle that case properly, and fallback to the old logic if that happens.

jtsiomb (changes requested):
`GENERATOR_IS_MULTI_CONFIG` is only available from cmake 3.9. If I'm reading the cmake docs correctly, older versions of cmake which don't have that property will just return `NOTFOUND` which I'm not sure what it does when fed to that `IF`. We should handle that case properly, and fallback to the old logic if that happens.

hwhsu1231:
@jtsiomb 

I posted a workaround compatible before CMake 3.9.

jtsiomb (approved):

jtsiomb:
Alright, this is clearly not a perfect solution, but it does the job while retaining backwards compatibilitiy to older versions of cmake, so I'll go ahead and merge it. Thank you.

PR #134, opened 2023-01-04
Add support for ARM64 MSVC targets

anthony-linaro:
Currently, the CMakeLists.txt file assumes CL_64 == x64, which means static lib builds (such as in blender) are broken on ARM64 target platforms. This makes a check for this, which allows the build to succeed.


jtsiomb:
Thank you!

PR #138, opened 2023-03-22
Fixes for Unicode support on Windows.

tweenk:
These changes should fix compiling freeglut with -DUNICODE -D_UNICODE on Windows.

### DIFF
--- a/src/mswin/fg_joystick_mswin.c
+++ b/src/mswin/fg_joystick_mswin.c
@@ -117,9 +117,11 @@ void fgPlatformJoystickRawRead( SFG_Joystick* joy, int* buttons, float* axes )
 
 static int fghJoystickGetOEMProductName ( SFG_Joystick* joy, char *buf, int buf_sz )
 {
-    char buffer [ 256 ];
+    TCHAR buffer [ 256 ];
+    const size_t bufferSize = sizeof buffer / sizeof buffer[0];
 
-    char OEMKey [ 256 ];
+    TCHAR OEMKey [ 256 ];
+    const size_t OEMKeySize = sizeof OEMKey / sizeof OEMKey[0];
 
     HKEY  hKey;
     DWORD dwcb;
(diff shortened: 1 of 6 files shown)


src/mswin/fg_window_mswin.c
```diff
@@ -661,35 +661,30 @@ void fgPlatformOpenWindow( SFG_Window* window, const char* title,
     }
 #endif /* !defined(_WIN32_WCE) */
 
+    TCHAR* tstr = fghTstrFromStr(title);
```
jtsiomb, line 664: For maximum compatibility, freeglut is written in C89. You can't define variables everywhere in the code, only at the top of a scope. Please move the delcaration of tstr to the start of the function.

src/mswin/fg_joystick_mswin.c
```diff
@@ -129,27 +129,27 @@ static int fghJoystickGetOEMProductName ( SFG_Joystick* joy, char *buf, int buf_
...
+                 REGSTR_PATH_JOYCONFIG, joy->pJoystick.jsCaps.szRegKey,
+                 REGSTR_KEY_JOYCURR );
 
     lr = RegOpenKeyEx ( HKEY_LOCAL_MACHINE, buffer, 0, KEY_QUERY_VALUE, &hKey);
 
     if ( lr != ERROR_SUCCESS ) return 0;
 
     /* Get OEM Key name */
-    dwcb = sizeof(OEMKey);
+    dwcb = _countof(OEMKey);
```
jtsiomb, line 141: _countof does not exist on MSVC 6. From the docs it looks like it's simply a macro expanding to `sizeof foo / sizeof *foo`, so please use that instead.

src/mswin/fg_joystick_mswin.c
```diff
@@ -129,27 +129,27 @@ static int fghJoystickGetOEMProductName ( SFG_Joystick* joy, char *buf, int buf_
...
+    _sntprintf ( buffer, _countof(buffer), _T("Joystick%d%s"), joy->pJoystick.js_id + 1, REGSTR_VAL_JOYOEMNAME );
 
     lr = RegQueryValueEx ( hKey, buffer, 0, 0, (LPBYTE) OEMKey, &dwcb);
     RegCloseKey ( hKey );
 
     if ( lr != ERROR_SUCCESS ) return 0;
 
     /* Open OEM Key from ...MediaProperties */
-    _snprintf ( buffer, sizeof(buffer), "%s\\%s", REGSTR_PATH_JOYOEM, OEMKey );
+    _snprintf ( buffer, _countof(buffer), _T("%s\\%s"), REGSTR_PATH_JOYOEM, OEMKey );
```
jtsiomb, line 152: shouldn't this also be `_sntprintf`?

jtsiomb (changes requested):
I'm not usually involved with the windows side of things, so I'll wait a while for anyone else to respond. I only have a couple of general comments for now.

src/mswin/fg_joystick_mswin.c
tweenk: Yes, fixed.

tweenk:
Uploaded new version with the requested changes

dcnieho:
yes, i can test once its ready

tweenk:
It looks like I made a typo when 'exporting' the change. For context, I do not have access to a Windows machine at the moment. I made these changes to a copy of the library in the Google internal monorepo, then compiled an internal demo application that uses freeglut using a cross-compilation toolchain and ran it in Wine. I then manually applied the same changes to a GitHub checkout and uploaded it, but I made a mistake during this process.

To ensure I didn't make this type of mistake this time, I used the automatic source copying tool [Copybara](https://github.com/google/copybara).

dcnieho:
Thanks! So this should be ready to test?

tweenk:
Yes, it should be ready to test now.

dcnieho:
Tested, compiles and the demo apps work fine

PR #139, opened 2023-03-30
Update README.cmake to cover building on recent versions of Visual Studio

aggserp4:
While trying to build FreeGLUT on MS Windows using Visual Studio, I noticed that README.cmake is outdated. The instructions for building on Windows involve installing CMake 2.8 where as 3.0.0 is referenced as the minimum in the README file. Also, CMake comes bundled with recent versions of VS meaning that installing it separately is unnecessary.


dcnieho:
Thanks for the contribution! FreeGLUT also aims to compatible with much older Visual Studio versions, so we can't simply replace this text. Could you update the pull request to make what you sent now an addition to the original text, and update the original text?

aggserp4:
I amended my commit to restore the original text with a note for those who may want to use the new method. Is this alright?

jtsiomb:
Thank you, this was forgotten for quite some time, but it looks fine to me, so I went ahead to merge it.

PR #146, opened 2023-09-26
properly respect user install dirs

mxraw:


jtsiomb:
As it is, this changeset seems to break MSVC builds, something about the location of the pdb file, and also increases the minimum cmake version number. Is it necessary to increase the version number? What does it buy us?

If part of this can be achieved without increasing the cmake version number (and obviously stops breaking the MSVC build) it would be easier to merge that first, and discuss the merits of the rest separately.

mxraw:
> As it is, this changeset seems to break MSVC builds,

Sorry, our build system configures CMake in a non-standard way and forgot to take that into account. I have fixed that issue.

> Is it necessary to increase the version number? What does it buy us?

Yes. 3.1 is what is called 'Modern CMake'. The changes from Version 3.0 are very numerous, fleshing out and/or completing what has been introduced in 3.0.
Among them proper target support and generator expressions.

Specifically, 3.1 is required for **[TARGET_PDB_FILE](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:TARGET_PDB_FILE)**

Personally, I don't recommend a version below 3.1.

jtsiomb:
TARGET_PDB_FILE seems to define where to place the msvc debug symbols file. What do we gain by being able to specify it? What was the previous behavior and what changes?

If being able to override the default location of the debug information file for MSVC is the only reason for bumping the required version, it doesn't sound enticing enough.

Just being able to use a different (better? more powerful?) syntax in the cmake file itself is not reason enough either, if we can achieve the same result both ways.

To bump the required cmake version there must be something very useful we gain by doing that. Something that cannot be achieved by other means.

mxraw:
> TARGET_PDB_FILE seems to define where to place the msvc debug symbols file. What do we gain by being able to specify it? What was the previous behavior and what changes?

The previous hand-crafted code is very brittle and easily points to the wrong location of the PDB, thus breaking the build. It won't even work on most MSVC installations beyond the 2 versions used in the CI here, especially when the packager modifies the standard locations.

The "new" (CMake 3.1 was released on December 17 2014) behavior using TARGET_PDB_FILE is the correct method to refer to the location of the PDB and should always just work, even if the packager decides to overwrite the default installation directories.

> If being able to override the default location of the debug information file for MSVC is the only reason for bumping the required version, it doesn't sound enticing enough.

It is very hard to impossible to get the correct location of the PDB file without TARGET_PDB_FILE, which is the reason it was introduced to CMake in the first place. The present code may be sufficient to generate a green checkmarks for the 2 CI builds here but it is wrong.

To go from hand-crafted specialty code that is wrong except corner cases to something that just works is well worth the 3.1 bump. This solves major headaches downstream.

The only downside to this change I see is that someone somewhere in the universe still runs the ancient CMake 3.0 and might have to upgrade to a slightly less ancient version.

> To bump the required cmake version there must be something very useful we gain by doing that. Something that cannot be achieved by other means.

As I said, retrieving the location of the PDB without TARGET_PDB_FILE is a very hard problem requiring lots of brittle hand-crafted code without 3.1.

Apart from that, 3.1 is very worth it and 3.0 and below should not be used anymore. To see why you will have to read up on the evolution of CMake yourself, it's not possible to summarize that here.

jtsiomb:
You'll have to be very specific about behaviors and effects I'm afraid. I'm not a windows user, I'm not an MSVC user, and I'm not a cmake user.

> The previous hand-crafted code is very brittle and easily points to the wrong location of the PDB, thus breaking the build. It won't even work on most MSVC installations beyond the 2 versions used in the CI here, especially when the packager modifies the standard locations.

The two versions in the CI are the extrema of "oldest MSVC version we want to support" and "newest MSVC version available". Is there some version in between for which the current behavrior breaks? And if so in what way? I don't remember ever seeing a bug report about this.

> It is very hard to impossible to get the correct location of the PDB file without TARGET_PDB_FILE, which is the reason it was introduced to CMake in the first place.

What happens if you get the location wrong? Does the install target fail? Linking should not need the PDB file unless I'm mistaken. Right?

> The only downside to this change I see is that someone somewhere in the universe still runs the ancient CMake 3.0 and might have to upgrade to a slightly less ancient version.

Another downside is that cmake 3.0.0 is the last version, as far as I can tell, which supports generating MSVC6 project files. Correct me if I'm wrong. If not then maybe bumping the version to 3.1 might not be such a huge deal indeed.

> Apart from that, 3.1 is very worth it and 3.0 and below should not be used anymore. To see why you will have to read up on the evolution of CMake yourself, it's not possible to summarize that here.

I'm not interested in cmake or its history. I'm just trying to determine exactly what we win and what we lose in either case to make an informed decision of whether it's worth it.

Is there a way to conditionally use the new feature if it's available, but fallback to the previous behavior if the user has cmake 3.0.0 ?

mxraw:
> What happens if you get the location wrong? Does the install target fail? 

Yep.

> Linking should not need the PDB file unless I'm mistaken. Right?

Yep. Unfortunately simply turning PDBs off isn't optional for my use case.

> Another downside is that cmake 3.0.0 is the last version, as far as I can tell, which supports generating MSVC6 project files.

MSVC6 was deprecated in 3.3 and removed in 3.6 according to the CMake docs.

> Edit: there's also the issue that on older systems it's often hard or impossible to build new versions of cmake at all, which is why I'd like to be conservative with what we require and keep it at the absolute minimum.

I understand that completely, believe me. CMake 3.1 is agreed to be the 'absolute minimum' since a couple of years though. If you buy any one of the books about it, it will tell you to use 'at least 3.1'.

> Is there a way to conditionally use the new feature if it's available, but fallback to the previous behavior if the user has cmake 3.0.0 ?

That is possible.

jtsiomb:
You're right, I booted up my windows machine and checked. cmake 3.1 still supports MSVC6 project generation. I'll take another look at your changes as soon as I'm able, and if there's nothing else objectionable I'll go ahead and merge it.

PR #155, opened 2024-02-12
Plug memory leak that happens upon error.

sebras:
If fgStructure.CurrentMenu is set when glutAddMenuEntry() or glutAddSubMenu() is called the allocated menuEntry variable will leak. This commit postpones allocating menuEntry until after the error checks, thereby plugging the memory leak.

This fixes CVE-2024-24258 and CVE-2024-24259.


sebras:
I don't know why these CVEs were filed against MuPDF instead of freeglut, and I doubt that these should be of high severity. Anyway, here is my proposed fix. Please let me know if you want me to change it some how.

jtsiomb:
Thanks! I've no idea how a memory leak can be classified as a "vulnerability", but nevertheless, it's good to not have any :)

sebras:
Me neither, but thank you for a speedy merge. :)

PR #159, opened 2024-04-18
Add "FGAPI" before all core instanciation  of Interface functions of Freeglut v3.4.0

tdechaize:
In addition of my tests of build Freeglut v3.4.0 on Windows, I constat that Pelles C Compiler is most very strict than all others compilers tested (CLANG,MINGW, MSVC, DMC, Borland C, etc...). It impose that all core instanciation of Interface functions of FG be preceed by FGAPI like declaration in "include\GL\freeglut_std.h".
For all others compilers tested this evolution is totally transparent, I realize this operation on all sources concerned of FG.
Thank's.
Thierry D.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,8 +11,126 @@ if (POLICY CMP0072)
     cmake_policy(SET CMP0072 NEW)
 endif()
 
+# NOTE :	Added at 2024/03/16, on Windows, you can choose many "free" compilers to generate Freeglut.
+#			All of these "free" compilers are installed on "specific" directory, same like MSVC (not in "/usr/bin" ... -))
+#			This addon, is mandatory to determine which compiler or devlopment environment is used and parameter this.
+#   WARNING : 		On my system, to choose "good" directories of binaries, include files and libraries, I decide to parameter
+#					many environment variables (system level) to point to these (because version's num defined directories ... GRRR) : 
+#									KIT_WIN_NUM  		valued at time 		10.0.22621.0
+#									KIT_WIN_VERSION  	valued at time 		10
+#									VS_NUM 				valued at time		14.39.33519
+#									VS_VERSION 			valued at time		2022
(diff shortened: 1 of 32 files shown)


jtsiomb (changes requested):
I started writing detailed comments, but it quickly became apparent that it's pointless. This is also a bunch of random things thrown together, not just the FGAPI fixes as your title suggests.

The CMakeLists.txt changes are a huge mess with grammatical mistakes in error messages, weird absolute paths, and a wall of text of comments. This is unacceptable.

For the rest of the changes, beyond the fact that you need to move anything other than FGAPI additions to separate pull requests:

  - drop unnecessary comments explaining the rationale for changes
  - drop dates of changes from comments
  - make sure to only use C89-style `/* .. */` comments
  - drop unnecessary formatting and whitespace changes.

jtsiomb:
Also, some of your changes are adding extra conditions to `HAVE_WHATEVER_H` for specific compilers. I feel like this is probably an issue in your machine. Why would cmake find the file, but the compiler fail to do so? Before you resubmit this kind of workarounds, try to reproduce the problem on a clean VM with only that specific compiler installed.

tdechaize:
Hi,
Yes, you rigth, and forgive me. I mix two versions on two different directories of very important file : CMakeLists.txt on last commit, with use of GitHub Desktop (newbie fault ... !!).
Today, I renew this commit whith many verifications. 
One add don't concern new prefix FGAPI in code interface functions of Freeglut version 3.4.0 : in file "freeglut_std.h", I conserve test on additional compilers on Windows about #define (declaration) of "bitmap" and "stroke" ident of characters set in this file (line 216). All tested with success.
I think that this add has not impact on rest of generation of FG (added ... || defined(__BORLANDC__) || defined(__DMC__) || defined(__LCC__) || defined(__POCC__))
Sorry for last commit, I learn about my errors ...
Thank's.

PR #160, opened 2024-04-25
Fix geometry drawing for OpenGL < 2.0

mardy:
The preprocessor guards are wrong: we do not want to compile this code unless OpenGL >= 2.0 is supported.

Incidentally, I have a question: I'm working on a FreeGLUT backend for the Nintendo GameCube and Nintendo Wii; would you accept this contribution (once it's ready), or is it better if it's kept in some fork (I could probably get it hosted by devkitPro)?

### DIFF
--- a/src/fg_geometry.c
+++ b/src/fg_geometry.c
@@ -404,7 +404,7 @@ static void fghDrawGeometryWire20(GLfloat *vertices, GLfloat *normals, GLsizei n
                                   GLushort *vertIdxs2, GLsizei numParts2, GLsizei numVertPerPart2,
                                   GLint attribute_v_coord, GLint attribute_v_normal)
 {
-#if defined(GL_VERSION_1_1) || defined(GL_VERSION_ES_CM_1_0)
+#if defined(GL_VERSION_2_0) || defined(GL_VERSION_ES_CM_1_0)
     GLuint vbo_coords = 0, vbo_normals = 0,
         ibo_elements = 0, ibo_elements2 = 0;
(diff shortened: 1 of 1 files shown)


jtsiomb:
No, this is incorrect for a couple of reasons:
  1. VBOs are in OpenGL core from version 1.5, not version 2.0.
  2. Many OpenGL 1.x less than 1.5 implementations still support VBOs through extensions. And we'd like to use them if they are available.
  3. Some OpenGL greater than 1.5 implementations, while they do have VBOs as core without the need for extension loading, they still don't define GL_VERSION_1_5 for compatibility reasons in their main gl.h header file.

Versions and availability of features need to be queried at runtime, and they are. Did you encounter a specific issue you're trying to fix with this? If so please open a bug report so we can discuss the best way to fix it.

About gamecube and nintendo wii support, I'm personally in favor of merging, as long as it doesn't need unreasonably intrusive changes. If you can add a new self-contained platform directory under src, similar to how we have `src/x11`, and `src/mswin`, add something relatively minimal to the cmake build file (or something in altbuild), and you're willing to keep an eye out for its maintainance for the forseeable future, I don't see why not.

mardy:
You are absolutely right. I had another wrong change in my code (I was defining `GL_ES_VERSION_2_0` in my cmake file to workaround another issue, which I now understand to be non-existant) and that was breaking `fg_geometry.c`. There's no issue, indeed.

Then stay tuned for the Wii/GC changes, sooner or later they'll arrive :-)

PR #167, opened 2024-05-07
Fix macOS actions

Fancy2209:
install xquartz and changes the libgl path, only way I found to fix it.


jtsiomb:
Thanks, merged manually to make some minor changes

PR #168, opened 2024-05-11
Wii mouse cursor, misc fixes

mardy:
Hi again :-)

Here's a few more changes for the Wii/GameCube port. The biggest one is the support for drawing the mouse cursor at the end of each frame, but there are other unrelated changes too.

Please let me know if you'd rather have different PRs for each issue; I thought, that since changes are all about this platform and are anyway split into different commits, it might not be hard to review this commit-by-commit, but you decide :-)


jtsiomb (changes requested):
I apologise for not mentioning this earlier, in your previous pull request, but I only now noticed you're using C99 features in these commits. The freeglut codebase is C89, so there's a number of things to keep in mind like declaring variables at the top of the scope, not having inline, bool, and sized types and so on.

On the cross-platform parts of freeglut these are usually caught by the build failing on MSVC6, but here you'll have to make a concsious effort to keep it C89-clean as much as possible.

We should look into the gcc diagnostics to see if it can help us with that at some point, but if I remember correctly passing -std=c89 to it, also disables large chunks of the libc headers for some unfathomable reason on GNU/Linux, and passing -std=gnu89 introduces all those GCC extensions which ended up as C99 anyway.

jtsiomb:
Oh an about the batching of commits, I think it's perfectly fine for this early stage of implementation where lots of things are happening. If you were in maintainance mode and attempting to fix bugs, then it would be better to have one commit per specific bug.

mardy:
Thanks John, I updated the branch.

> We should look into the gcc diagnostics to see if it can help us with that at some point, but if I remember correctly passing -std=c89 to it, also disables large chunks of the libc headers for some unfathomable reason on GNU/Linux, and passing -std=gnu89 introduces all those GCC extensions which ended up as C99 anyway.

Once opengx will be packaged in devkitPro I can submit a PR which adds CI for the Wii/GameCube, we do have them in many projects (like in opengx itself: https://github.com/devkitPro/opengx/actions).

jtsiomb:
That would be great, but I wasn't thinking just about CI, but rather in general. If when you're building freeglut it warned you about C89 compliance it would help keep things straight when building without a strictly C89 compiler.

PR #181, opened 2024-08-10
fix building problems of demos when enable FREEGLUT_REPLACE_GLUT option

CreativeLau:
Just fix some building problems for demos once the FREEGLUT_REPLACE_GLUT option is enabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,11 @@ ELSE()
     OPTION(FREEGLUT_REPLACE_GLUT "Be a replacement for GLUT" ON)
 ENDIF()
 
+# determine link glut or freeglut in freeglut_std.h
+IF(FREEGLUT_REPLACE_GLUT)
+    ADD_DEFINITIONS(-DFREEGLUT_REPLACE_GLUT)
+ENDIF()
+
 IF(NOT WIN32)
     # Wayland support
     OPTION(FREEGLUT_WAYLAND "Use Wayland (no X11)" OFF)
@@ -538,7 +543,7 @@ IF(WIN32)
     IF(FREEGLUT_BUILD_STATIC_LIBS)
         TARGET_COMPILE_DEFINITIONS(freeglut_static PUBLIC FREEGLUT_STATIC)
         IF(FREEGLUT_REPLACE_GLUT)
-            SET_TARGET_PROPERTIES(freeglut_static PROPERTIES OUTPUT_NAME ${LIBNAME})
+            SET_TARGET_PROPERTIES(freeglut_static PROPERTIES OUTPUT_NAME ${LIBNAME}_static)
         ENDIF()
         # need to set machine:x64 for linker, at least for VC10, and
         # doesn't hurt for older compilers:
--- a/include/GL/freeglut_std.h
+++ b/include/GL/freeglut_std.h
@@ -70,10 +70,18 @@
 
         /* Link with Win32 static freeglut lib */
 #       if FREEGLUT_LIB_PRAGMAS
-#           ifdef NDEBUG
-#              pragma comment (lib, "freeglut_static.lib")
+#           ifdef FREEGLUT_REPLACE_GLUT
+#               ifdef NDEBUG
+#                   pragma comment (lib, "glut_static.lib")
+#               else
+#                   pragma comment (lib, "glut_staticd.lib")
+#               endif
 #           else
-#              pragma comment (lib, "freeglut_staticd.lib")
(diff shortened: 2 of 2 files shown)


jtsiomb:
Changing freeglut_std.h to honor a define available during the build of freeglut is extremely pointless.

About the second change, making the output of the static build on windows be glut_static.lib, can you give us a little more information? Was it previously overwriting the dll import library? What are you trying to achieve?

CreativeLau:
compile envrionment:
CMake 3.26.3
VS2019 19.29
Windows SDK version 10.0.22621.0

If the option FREEGLUT_REPLACE_GLUT is off, there is no problem building demos e.g. 3dview 3dview_static.

If trun on the option FREEGLUT_REPLACE_GLUT, I got 2 problems during building demos e.g. 3dview 3dview_static.

- First, there is a link error as following:

``` shell
# 3dview
Build started...
1>------ Build started: Project: 3dview, Configuration: Release x64 ------
1>LINK : fatal error LNK1104: cannot open file 'freeglut.lib'
1>Done building project "3dview.vcxproj" -- FAILED.
========== Build: 0 succeeded, 1 failed, 2 up-to-date, 0 skipped ==========

# 3dview_static
Build started...
1>------ Build started: Project: 3dview_static, Configuration: Release x64 ------
1>LINK : fatal error LNK1104: cannot open file 'freeglut_static.lib'
1>Done building project "3dview_static.vcxproj" -- FAILED.
========== Build: 0 succeeded, 1 failed, 2 up-to-date, 0 skipped ==========
```

The reason is the pragma in freeglut_std.h link freeglut.lib but the lib name has been changed to glut.lib in CMakeLists.

``` c
# freeglut_std.h line 65-100

/* Windows static library */
#   ifdef FREEGLUT_STATIC

#       define FGAPI
#       define FGAPIENTRY

        /* Link with Win32 static freeglut lib */
#       if FREEGLUT_LIB_PRAGMAS
#           ifdef NDEBUG
#              pragma comment (lib, "freeglut_static.lib")
#           else
#              pragma comment (lib, "freeglut_staticd.lib")
#           endif
#       endif

/* Windows shared library (DLL) */
#   else

#       define FGAPIENTRY __stdcall
#       if defined(FREEGLUT_EXPORTS)
#           define FGAPI __declspec(dllexport)
#       else
#           define FGAPI __declspec(dllimport)

            /* Link with Win32 shared freeglut lib */
#           if FREEGLUT_LIB_PRAGMAS
#               ifdef NDEBUG
#                   pragma comment (lib, "freeglut.lib")
#               else
#                   pragma comment (lib, "freeglutd.lib")
#               endif
#           endif

#       endif

#   endif
```

``` cmake
# CMakeLists.txt line 529-531
IF(FREEGLUT_REPLACE_GLUT)
    SET(LIBNAME glut)
ENDIF()
```

To sovle this issue, I add the definition FREEGLUT_REPLACE_GLUT in CMakeLists and if...else to choose link freeglut or glut library.

- Sencond, both of the name of shared and static lib is glut.lib, but they should be different. If run Install, there is only one glut.lib of shared library. The static library was overwrite. The static library name should add a postfix _static as glut_static, just like freeglut_static.

``` cmake
# CMakeLists.txt 540-542
IF(FREEGLUT_REPLACE_GLUT)
    SET_TARGET_PROPERTIES(freeglut_static PROPERTIES OUTPUT_NAME ${LIBNAME})
ENDIF()
```

jtsiomb:
The problem with adding a define during build is that it might solve the issue of building demos, but it's an addition to a public header that doesn't do anything when used by other programs. That pragma will still try to link with freeglut.lib regardless of whether the user decided to build with or without the replacement option.

I have a suggestion which will solve both problems simultaneously: drop the replace option on windows, it's pointless anyway because on windows the original GLUT was called glut32.lib anyway, not glut.lib, so we're not even replacing anything.

Alternatively if we want to make the replacement feature work on windows, we should change the output to glut32.lib, and drop the static build when it's defined, so that it doesn't overwrite it. Then *generate* a new header file with the pragma in it and include it from freeglut_std.h, so that it links the correct library.

I won't act on this one way or another until I hear from the windows people around here. Meanwhile we could attempt to fix the demo build by linking them explicitly with freeglut instead of relying on that stupid pragma, and define FREEGLUT_LIB_PRAGMAS to 0 when building demos. I would merge a pull request fixing just this issue in this manner if you're interested in submitting it.

CreativeLau:
I argree with your suggestion definitely. I will reconsider this issue by changing the demo's configurations.

PR #184, opened 2024-10-30
freeglut_std.h: don't use X11 fonts when building with devkitPro

mardy:
This fixes an issue uncovered when building an application using the GLUT_BITMAP_* constants under the Nintendo Wii: freeglut_std.h was defining these constants to point to the variables defined in the X11 font module.

We fix this by checking for the presence of the __DEVKITPRO__ macro, which is a macro always defined when using the GCC built by the devkitPro distribution. Since this is a distribution meant for gaming consoles, we can be sure that X11 is not available there, so let's use the fonts embedded in freeglut.

### DIFF
--- a/include/GL/freeglut_std.h
+++ b/include/GL/freeglut_std.h
@@ -217,7 +217,7 @@
  *
  * Steve Baker suggested to make it binary compatible with GLUT:
  */
-#if defined(_MSC_VER) || defined(__CYGWIN__) || defined(__MINGW32__) || defined(__WATCOMC__)
+#if defined(_MSC_VER) || defined(__CYGWIN__) || defined(__MINGW32__) || defined(__WATCOMC__) || defined(__DEVKITPRO__)
 #   define  GLUT_STROKE_ROMAN               ((void *)0x0000)
 #   define  GLUT_STROKE_MONO_ROMAN          ((void *)0x0001)
 #   define  GLUT_BITMAP_9_BY_15             ((void *)0x0002)


jtsiomb:
How are fonts handled in the nintendo codebase? Couldn't you just provide the same symbols that X11 defines for fonts?

the nintendo stuff being new, and there being nothing to be compatible with, you can certainly choose the X11 way with the symbols or the win32 way with the numbers, and I don't particularly care which, it's up to you, I'd just like to learn what your reasoning was.

All else being equal, and if there's no specific reason to pick this or that way, I'd personally prefer a solution which doesn't touch public headers, although again this being a new platform, the incentive isn't particularly strong.

mardy:
> How are fonts handled in the nintendo codebase? Couldn't you just provide the same symbols that X11 defines for fonts?

I need some help in understanding the code: I've been staring at `src/x11/fg_glutfont_definitions_x11.c` for a while, and I still have no clue about what that file does :-)
I was thinking that it's goal was to avoid defining the font glyphs in freeglut and instead use the fonts provided by X11, but I cannot find the place where X11 fonts are loaded and I cannot see an `#ifdef` aroung the glyphs definitions, so it looks like freeglut is defining its own glyphs even under X11.

> All else being equal, and if there's no specific reason to pick this or that way, I'd personally prefer a solution which doesn't touch public headers, although again this being a new platform, the incentive isn't particularly strong.

The way I see it, is that the nintendo backend is not the exception here, but the rule, along with all other platforms supported by freeglut, except for X11 and Wayland: these two are the real exceptions here, and it seems to me that the logic in the header file is currently reversed: it now checks for the compiler, to ensure that it's not a compiler used on a platform where one usually has X11 or Wayland, whereas the proper logic IMHO would be:

```c
#ifdef HAS_X11_OR_WAYLAND
// Use X11 fonts
#else
#   define  GLUT_STROKE_ROMAN               ((void *)0x0000)                                                                          
#   define  GLUT_STROKE_MONO_ROMAN          ((void *)0x0001)                                                                          
#   define  GLUT_BITMAP_9_BY_15             ((void *)0x0002)                                                                          
#   define  GLUT_BITMAP_8_BY_13             ((void *)0x0003)                                                                          
#   define  GLUT_BITMAP_TIMES_ROMAN_10      ((void *)0x0004)                                                                          
#   define  GLUT_BITMAP_TIMES_ROMAN_24      ((void *)0x0005)                                                                          
#   define  GLUT_BITMAP_HELVETICA_10        ((void *)0x0006)                                                                          
#   define  GLUT_BITMAP_HELVETICA_12        ((void *)0x0007)                                                                          
#   define  GLUT_BITMAP_HELVETICA_18        ((void *)0x0008)                                                                          
#endif
```

I guess the reason why it wasn't done in this way is because there's no easy way to come up with a definition "HAS_X11_OR_WAYLAND"?

jtsiomb:
I'm not very familiar with the font code. I'll look into it and get back at you. Original GLUT used `glXUseXFont`, it's documented in Kilgard's "green book", but freeglut I think uses the font data in `src/fg_font_data.c` for all platforms if I'm not mistaken. The way the header file declares all the fonts in freeglut was done to remain backwards compatible with original GLUT, because that was how *it* was declaring them too. And that's probably true for both schemes, so I would guess Nate Robbins' GLUT win32 port probably uses those numbers we're using under windows.

> The way I see it, is that the nintendo backend is not the exception here, but the rule, along with all other platforms supported by freeglut, except for X11 and Wayland

I don't think so. All the ifdefs I see are specifically for windows compilers, aside from watcom which I'm pretty sure is a bug, because watcom can also build for GNU/Linux. Unless I'm missing something, only windows uses the "small integers cast to void* scheme".

jtsiomb:
I completely forgot this. Anyway there's no point in delaying merging this just to decide which one of the two declarations is preferable for a new platform. I'll merge your choice of using the windows-style font indices, and if you change your mind before the next release of freeglut, feel free to change it. After release it will be set in stone for ever :)

PR #197, opened 2025-02-21
macOS cocoa stub

drewwoods:
- Provides an empty stub implementation for Cocoa to allow successful
  build and linking:
  - Located in `src/cocoa/*`
  - Used only when `FREEGLUT_COCOA=ON` (`cmake -DFREEGLUT_COCOA=ON`);
    otherwise, the default remains XQuartz on macOS.
  - Functions only call `fgWarning()` and return.

- ~~Added `src/cocoa/.clang-format` to match the existing code style as
  closely as possible.~~ We decided to keep `src/cocoa/.clang-format` out of the repo

- Updated `src/fg_internal.h` to include `src/cocoa/fg_internal_cocoa.h`
  when built with `FREEGLUT_COCOA=ON`.


src/cocoa/.clang-format
jtsiomb, line 1: Let's drop this file from the commit. There's no reason to include .clang-format in the freeglut distribution.

src/cocoa/fg_internal_cocoa.h
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+ * PAWEL W. OLSZTA BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#pragma once
```
jtsiomb, line 20: use the standard ifndef/define guard, for consistency with the rest of the code.

jtsiomb (changes requested):
Very good. Only a couple of minor changes and it's good to merge. When the implementation is fleshed out, I'd like to change into C source files any of these files which are not going to have to call any ObjC APIs, and keep as ObjC only those which have to be ObjC. But we can do that at the end, when the dust clears.

src/cocoa/fg_internal_cocoa.h
drewwoods: no problem

drewwoods (commented):
> I'd like to change into C source files any of these files which are not going to have to call any ObjC APIs, and keep as ObjC only those which have to be ObjC. But we can do that at the end, when the dust clears.

Sounds good, I'll keep that in mind as I roll out the implementation

jtsiomb (approved):
Perfect. But it can't be automatically merged. Do you want me to go ahead and rebase from the command-line, or would you rather rebase and update the pull request?

drewwoods:
> Perfect. But it can't be automatically merged. Do you want me to go ahead and rebase from the command-line, or would you rather rebase and update the pull request?

Ok, should be mergeable now, let me know if not.  I'm also fine if you find it easier to fix things on your end if there are still conflicts.

Issue #195, opened 2025-02-20
Implement Native macOS (Cocoa) Support


drewwoods:
## Overview
Working on implementing native macOS support for FreeGLUT using Cocoa and NSOpenGL. This will allow FreeGLUT to run natively on macOS without requiring XQuartz, while providing additional functionality not available in the default macOS GLUT implementation.

See XQuartz GLX issues below as a motivation for Cocoa implementation

## Benefits
- Eliminates XQuartz dependency for macOS users
- Provides native performance and better system integration
- Adds support for features missing from Apple's GLUT implementation:
  - Mouse scroll wheel support
  - Extended window management capabilities
  - Modern input device handling
  - Fuller GLUT API compatibility

## TODO

**1. Maintenance:**
  - [x] All demo applications build  
    - #192 
  - [x] All demo applications run 
    - #199
    - #201 
  - [ ] All demos match expected behavior

**2. Infrastructure:**  
   - [x] CMake configuration for Cocoa 
     - #196
   - [x] Basic header files and stubs 
     - #197

**3. Core Implementation:**   
   - [x] Window/Display handling 
     - #198 
     - #213 
     - #235
     - #236
   - [x] State management 
     - #198 
     - #219
     - #259
   - [x] Main loop integration 
     - #198 

**4. Features:**
   - [x] Input handling (keyboard, mouse)  
     - #198 
     - #215
     - #267
     - #270
   - [x] Menu system integration
      - #280 
   - [ ] Game mode support
   - [x] Subwindows
     - #272
     - #274 
     - #275 
   - [ ] Configurable mouse cursor image
   - [x] Window management features and Full screen support
     - #235

## Technical Details/Approach
- Using Cocoa's NSWindow and NSOpenGLView for window management
- Supporting modern OpenGL context creation or legacy OpenGL 2.1 profile
- Maintaining compatibility with existing FreeGLUT API

## Current Progress
Most of the core Cocoa functionality has now been implemented. See Known Limitations below for remaining gaps.

## Known Limitations
- Menu system not yet implemented  
- Game mode not yet implemented

## macOS VSync Stability 
[...]

PR #198, opened 2025-03-07
macOS: Core implementation of window management and event handling

drewwoods:
This PR provides a functional implementation of the Cocoa backend for
FreeGLUT on macOS.

Changes include:

 - Implemented OpenGL context creation and window management
 - Added NSWindow and NSOpenGLView classes with event handling
 - Added mouse and keyboard input handling with proper coordinate mapping
 - Implemented window delegate for handling window events
 - Added support for both OpenGL Legacy (≤2.1) and Core profiles (3.2-4.1)
 - Renamed several Objective-C (.m) files to C (.c) where appropriate
 - Added proper debug macros and improved error reporting
 - Implemented various platform functions like swap buffers, DLOpen for GL function loading
 - VSync support (partially implemented with CVDisplayLink)

Implemented fg functions:

 - fgPlatformInitialize: Initializes Cocoa application with screen detection
 - fgPlatformSystemTime: Gets system time via clock_gettime_nsec_np
 - fgPlatformGlutSwapBuffers: Buffer swapping implementation (partial, VSync pending)
 - fgPlatformGetProcAddress: OpenGL function loader using dlsym
 - fgPlatformGetGLUTProcAddress: GLUT function address lookup
 - fgPlatformOpenWindow: Window creation with pixel format configuration
 - fgPlatformCloseWindow: Window cleanup and destruction
 - fgPlatformSleepForEvents: Event waiting via NSApp nextEventMatchingMask
 - fgPlatformProcessSingleEvent: Event processing
 - fgPlatformDestroyContext: OpenGL context cleanup
 - fgPlatformSetWindow: Current OpenGL context management
 - fgPlatformReshapeWindow: Window resizing
 - fgPlatformCloseDisplay: Display cleanup
 - fgPlatformDeinitialiseInputDevices: Input device cleanup
 - fgPlatformMainLoopPreliminaryWork: App launch completion
 - fgPlatformCreateWindow: Window structure initialization

Testing Performed:

Basic testing was conducted to verify that the sample applications
function correctly with the Cocoa backend implementation. The tested
applications showed behavior consistent with their XQuartz counterparts,
except for features that use functionality not yet implemented in this
commit.

Below is a table of the current status of each of the sample applications / demos

| Demo               | Status                | Details                                                           |
|--------------------|----------------------|-------------------------------------------------------------------|
| 3dview             | Functional           | Menu functionality and mouse drag operations working properly      |
| CallbackMaker      | Partially Functional | Limited functionality as subwindows are not yet implemented        |
| Fractals           | Functional           | All features working as expected                                   |
| Fractals_random    | Functional           | All features working as expected                                   |
| Lorenz             | Not Functional       | Application crashes due to OpenGL usage before initialization      |
| One                | Partially Functional | Limited functionality as game mode is not yet implemented          |
| indexed_color      | Unsupported          | Index color mode not supported on macOS                            |
| joystick           | Not Functional       | Application exits during initialization as joystick support is pending |
| keyboard           | Functional           | Key, special, and modifier callbacks verified working              |
| multi-touch        | Partially Functional | Limited by unimplemented multibutton functionality                 |
| resizer            | Partially Functional | Position/resize/z-order operations not fully implemented           |
| shapes             | Functional           | All features working as expected                                   |
| smooth_opengl3     | Partially Functional | Requires modification for macOS OpenGL 3.2-4.1 compatibility       |
| spaceball          | Partially Functional | Limited as spaceball functionality is not implemented              |
| subwin             | Partially Functional | Subwindows display as independent windows                          |
| timer              | Functional           | All timer features working as expected                             |
| timer_callback     | Functional           | All timer callback features working as expected                    |

Progress on #195


include/GL/freeglut_std.h
```diff
@@ -141,8 +141,12 @@
...
 #       define GL_SILENCE_DEPRECATION
 #   endif
-#   include <OpenGL/gl.h>
-#   include <OpenGL/glu.h>
+/* don't include compatibility OpenGL headers if we have included OpenGL 3 headers */
+#   ifndef __gl3_h_
+#       include <OpenGL/gl.h>
+        /* GLU on MacOS is only supported with compatibility OpenGL headers */
+#       include <OpenGL/glu.h>
+#   endif
```
jtsiomb, line 149: is there a conflict if both are included? Because this seems rather fragile. It breaks if the user includes gl3 headers *after* freeglut.h, which would otherwise be the recommended order for other platforms. It also breaks if apple changes the include guard (which admittedly is less likely).

If there is indeed a conflict, then I think it would be better to have an explicit `FREEGLUT_NO_GL_INCLUDE` macro, or something like that, that can be defined by an application to avoid GL includes on any platform.

src/cocoa/fg_ext_cocoa.c
```diff
@@ -19,17 +19,37 @@
...
-
-    return NULL;
+    static void *glHandle = NULL;
+    if ( glHandle == NULL ) {
+        glHandle = dlopen( "/System/Library/Frameworks/OpenGL.framework/OpenGL", RTLD_LAZY | RTLD_GLOBAL );
+        if ( glHandle == NULL ) {
+            fgError( "Failed to dlopen OpenGL framework" );
+        }
+    }
+    void *addr = dlsym( glHandle, procName );
```
jtsiomb, line 53: move declarations before code (C89)

jtsiomb (changes requested):
This looks pretty good. I left a couple of specific comments about things I noticed. 

Beyond that, make sure that at least C source files and header files remain valid C89 code. That means declarations at the top before any code, and C-style comments `/* ... */` instead of C++-style comments '// ...'

I don't know what obj-c allows, so I'll leave that up to you.

drewwoods:
> This looks pretty good. I left a couple of specific comments about things I noticed.
> 
> Beyond that, make sure that at least C source files and header files remain valid C89 code. That means declarations at the top before any code, and C-style comments `/* ... */` instead of C++-style comments '// ...'
> 
> I don't know what obj-c allows, so I'll leave that up to you.

Sure no problem, how are you with variadic-macros?  When compiling under gnu89 its the only warnings I get from the c/h cocoa files

```sh
$ FILETYPE='cocoa\.c\|cocoa\.h'

$ export CFLAGS="-DGL_SILENCE_DEPRECATION -O0 -ggdb  -std=gnu89  -Wall -Wpedantic -fsanitize=address"
$ cmake -DCMAKE_BUILD_TYPE=Debug -DFREEGLUT_COCOA=ON ..

$ make 2>&1 | \
grep "$FILETYPE" | \
grep -o "\[\-W.*\]" | \
sort | \
uniq -c

 152 [-Wvariadic-macros]
```

I use them quite extensively for the debug logging:

```c
#ifdef DEBUG_LOG
#define DBG( ... ) printf( __VA_ARGS__ )
#else
#define DBG( ... )
#endif

#ifdef UNIMPLEMENTED_WARNING
#define PRINT_ONCE( ... )             \
    do {                              \
        static int once;              \
        if ( !once ) {                \
            fgWarning( __VA_ARGS__ ); \
            once = 1;                 \
        }                             \
    } while ( 0 )
#define TODO_IMPL PRINT_ONCE( "%s not implemented yet in Cocoa", __func__ )
#define PART_IMPL PRINT_ONCE( "%s partially implemented in Cocoa", __func__ )
#define NO_IMPL   PRINT_ONCE( "%s not implemented in Cocoa", __func__ )
#else
#define TODO_IMPL
#define PART_IMPL
#define NO_IMPL
#endif
```

If you prefer I can do something like:

```c
#include <stdio.h>
    
#define LOG(args) (printf("LOG: %s:%d ", __FILE__, __LINE__), printf args)
    
int main(void) {
    int n = 42; 
    LOG(("Hello, world\n"));
    LOG(("n = %d\n", n));
    return 0;
}   
```

https://stackoverflow.com/questions/27663053/variadic-macros-alternative-in-ansi-c

### Obj-C warnings with -std=gnu89

```sh
$ FILETYPE='cocoa\.m\|cocoa\.h'
$ export CFLAGS="-DGL_SILENCE_DEPRECATION -O0 -ggdb  -std=gnu89  -Wall -Wpedantic -fsanitize=address"
$ cmake -DCMAKE_BUILD_TYPE=Debug -DFREEGLUT_COCOA=ON ..

$ make 2>&1 | \
grep "$FILETYPE" | \
grep -o "\[\-W.*\]" | \
sort | \
uniq -c

  26 [-Wdeclaration-after-statement]
   2 [-Wlong-long]
 152 [-Wvariadic-macros]
```

include/GL/freeglut_std.h
drewwoods: > is there a conflict if both are included? Because this seems rather fragile

Ye sadly, here is

`OpenGL.framework/Versions/A/Headers/gl.h`
```c
     1 #ifndef __gl_h_                                                                                                                                                    
     2 #define __gl_h_                                                                                                         
     3                                                                                                                         
     4 #if defined __gl3_h_ && !(defined GL_DO_NOT_WARN_IF_MULTI_GL_VERSION_HEADERS_INCLUDED)                                  
     5 #warning gl.h and gl3.h are both included.  Compiler will not invoke errors if using removed OpenGL functionality.      
     6 #endif    
```

`OpenGL.framework/Versions/A/Headers/gl3.h`
```c
     1 #ifndef __gl3_h_                                                                                                                                                   
     2 #define __gl3_h_                                                                                                        
     3                                                                                                                         
     4 /* Allow multiple GL version headers to be included when we're working                                                  
     5    with modules. */                                                                                                     
     6 #ifdef __has_feature                                                                                                    
     7 #  if __has_feature(modules)                                                                                            
     8 #    ifndef GL_DO_NOT_WARN_IF_MULTI_GL_VERSION_HEADERS_INCLUDED                                                         
     9 #      define GL_DO_NOT_WARN_IF_MULTI_GL_VERSION_HEADERS_INCLUDED 1                                                     
    10 #    endif                                                                                                              
    11 #  endif                                                                                                                
    12 #endif                                                                                                                  
    13                                                                                                                         
    14 #if defined __gl_h_ && !(defined GL_DO_NOT_WARN_IF_MULTI_GL_VERSION_HEADERS_INCLUDED)                                   
    15 #warning gl.h and gl3.h are both included.  Compiler will not invoke errors if using removed OpenGL functionality.      
    16 #endif                                                                                                                  
    17                 
```

> If there is indeed a conflict, then I think it would be better to have an explicit FREEGLUT_NO_GL_INCLUDE macro

Ahh ye that sounds better, let me do that

drewwoods: Ok I've updated the code with a code guard

https://github.com/freeglut/freeglut/pull/198/files#diff-278200a05de757c016a8c1e45e7ad3486638b191f2b0e09e9a6b916c45b28049L130-L145

For an example of how this is used, please see PR #199

jtsiomb:
> It has the same warnings as C. Objective-C is a strong superset of C, unlike C++. So the C parts of objective-c with `-std=gnu89` will have the same adherence as pure C.
> 
> That being said, unlike C89 being a valid usecase for older platforms, the obj-C code here would not be backward compatible. The `-Wdeclaration-after-statement` etc warnings will be the least of your worries if you attempting to compile on an old obj-c compiler–– so I dont think its worth making the modifications for the obj-c code, unless you have a strong preference stylistically.

For the moment I don't have a strong preference when it comes to the obj-c files. I'll leave it up to your judgment. I just want header files which can be included from C, and C source files to be C89-compliant (and anything newer in ifdefs with appropriate fallbacks for C89).

But if you can make the code compatible with older macosx versions, as far back as it's practical to go, I would consider that a worthy goal. Of course we can leave that for future improvements if you don't feel like it. I don't consider it a show-stopper if you only decide to support the latest macosx for the first incarnation of this new backend. So again, I'll leave it up to your judgment.

src/cocoa/fg_ext_cocoa.c
drewwoods: Sure, done

drewwoods:
@jtsiomb thanks for the feedback, I've addressed all your comments, the pure C portions are now C89 compliant.

I address the following:
 - declaration-after-statement
 - comments // --> /**/
 - remove variadic-macros

PR #201, opened 2025-03-07
macOS: Update Lorenz demo to run on macOS (Cocoa)

drewwoods:
Only call OpenGL functions after glutInit/glutCreateWindow.

In Cocoa the OpenGL context is only created during windows creation (glutCreateWindow)

### DIFF
--- a/progs/demos/Lorenz/lorenz.c
+++ b/progs/demos/Lorenz/lorenz.c
@@ -341,11 +341,6 @@ int main ( int argc, char *argv[] )
   /* Initialize the random number generator */
   srand ( 1023 ) ;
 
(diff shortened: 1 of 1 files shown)


jtsiomb:
> Only call OpenGL functions after glutInit/glutCreateWindow.
> 
> In Cocoa the OpenGL context is only created during windows creation (glutCreateWindow)

That's true on every platform. The 3 calls were silently ignored by other OpenGL implementations. clear color 0,0,0,1 is the default, clear depth 1 is the default, and the demo never managed to enable depth testing before, but since it's only drawing lines, nobody noticed.

PR #202, opened 2025-03-10
build: remove strict-prototype warnings

drewwoods:
Removes these types of warnings:
 - `warning: a function declaration without a prototype is deprecated in all versions of C [-Wstrict-prototypes]`

Test builds ran:
 - `CFLAGS="-Wall -Wpedantic"  cmake -DCMAKE_BUILD_TYPE=Debug .. && make clean && make`
 - `CFLAGS="-std=gnu89 -Wall -Wpedantic" cmake -DCMAKE_BUILD_TYPE=Debug .. && make clean && make`

### DIFF
--- a/include/GL/freeglut_ext.h
+++ b/include/GL/freeglut_ext.h
@@ -214,7 +214,7 @@ FGAPI void    FGAPIENTRY glutSolidTeaspoon( double size );
 /*
  * Extension functions, see fg_ext.c
  */
-typedef void (*GLUTproc)( void );
+typedef void (*GLUTproc)();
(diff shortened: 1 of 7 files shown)


jtsiomb (changes requested):
Modifying the function pointer signature in a public header is not something I want to do without a much stronger reason than getting rid of a warning. This will introduce an API incompatibility with previous versions, which is not acceptable.

Every other change in internal files is absolutely fine, but let's not make incompatible changes in `include/`

drewwoods:
> Every other change in internal files is absolutely fine, but let's not make incompatible changes in `include/`

Good point and well caught

PR #204, opened 2025-03-14
demos: Add vsync demo to test proper double buffering

drewwoods:
This commit adds a new demo application that tests proper vsync implementation with double buffering. The program draws a single block per frame without clearing the buffer, creating a checkerboard pattern that alternates between two color sets.

In a correctly synchronized double-buffered environment:
- The screen should completely fill with blocks
- The two buffers will each contain half the blocks (in a checkerboard pattern)
- Colors will appear to flash rapidly as the buffers alternate
- Frame rate should match the monitor's refresh rate

In environments with improper vsync observe:
- Missing blocks
- Only one of the two buffers may be visible
- Inconsistent rendering across the screen

This visual test makes it easy to identify proper vsync behavior versus broken implementations.

This test was used in the development of the Cocoa implementation

### vsync demo with #203 applied:

https://github.com/user-attachments/assets/bf439b15-634e-420f-bc76-c708175eeafb

### vsync demo prior to #203:

https://github.com/user-attachments/assets/86f83dad-e1fa-4c76-b1b7-379010f675bb

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -759,6 +759,7 @@ ADD_DEMO(timer_callback  progs/demos/timer_callback/timer.c)
 ADD_DEMO(keyboard        progs/demos/keyboard/keyboard.c)
 ADD_DEMO(indexed_color   progs/demos/indexed_color/idxcol.c)
 ADD_DEMO(3dview          progs/demos/3dview/3dview.c)
+ADD_DEMO(vsync           progs/demos/vsync/vsync.c)
 
 #######################
 # INSTALLATION
--- a/progs/demos/vsync/vsync.c
+++ b/progs/demos/vsync/vsync.c
@@ -1,0 +1,137 @@
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+#include <time.h>
+
+#include <GL/freeglut.h>
+
+#define ARR_LEN( arr ) ( sizeof( arr ) / sizeof( arr[0] ) )
+
+static uint64_t last_switch;
+static uint32_t frames;
+static char     buf[1024] = "VSync test:\n"
+                            "===========\n"
+                            "In a correctly synchronized double-buffered environment:\n"
+                            " - The screen should completely fill with blocks\n"
+                            " - The two buffers will each contain half the blocks (in a checkerboard pattern)\n"
+                            " - Colors will appear to flash rapidly as the buffers alternate\n"
+                            " - Frame rate should match the monitor's refresh rate\n"
+                            "\n"
+                            "In environments with improper vsync you'll observe:\n"
+                            " - Missing blocks\n"
+                            " - Only one of the two buffers may be visible\n"
+                            " - Inconsistent rendering across the screen\n "
+                            "\n";
+
+static char *msg = NULL;
+
+static int   idx          = 0;
+static float colours[][3] = {
+    { 0.6f, 0.0f, 0.0f }, /* Red */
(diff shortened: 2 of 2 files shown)


jtsiomb:
While I'd like to keep this test program around, the freeglut demos are not a suite of regression tests, they are demonstrating some aspect of using freeglut, they are mainly example programs.

Since freeglut doesn't have any feature for controlling vsync, this is not really demonstrating anything, and it's basically a tossup on how it'll behave on any other platform where there is some implementation-specific default vsync state and you need to control it with an GLX/WGL extension in the program.

On the other hand, this might be the perfect opportunity to actually add vsync control to freeglut, and make this test program use it. It is an important thing that needs to be controlled in a platform-specific way, so I'd say it falls squarely within the freeglut scope.

dcnieho:
+1 this is a good idea

drewwoods:
> While I'd like to keep this test program around, the freeglut demos are not a suite of regression tests, they are demonstrating some aspect of using freeglut, they are mainly example programs.

I agree that it’s clearer if we keep developer tests separate from user samples. Would it be worthwhile to create a separate repository for development and regression tests?

> Since freeglut doesn't have any feature for controlling vsync, this is not really demonstrating anything, and it's basically a tossup on how it'll behave on any other platform where there is some implementation-specific default vsync state and you need to control it with an GLX/WGL extension in the program.

Ah, I had mistakenly assumed that double buffering automatically implied vsync, which I discovered through empirical testing with GLX FreeGLUT applications.

> On the other hand, this might be the perfect opportunity to actually add vsync control to freeglut, and make this test program use it. It is an important thing that needs to be controlled in a platform-specific way, so I'd say it falls squarely within the freeglut scope.

Makes sense, I opened tracking issue #206

jtsiomb:
> I agree that it’s clearer if we keep developer tests separate from user samples. Would it be worthwhile to create a separate repository for development and regression tests?

I don't think it's worthwhile if it will only have a single regression test :) It might be useful if we were to create a whole suite of tests, and have them run automatically like the build testing we currently have as github actions, but that's a lot of work. Also I'm not convinced that automated testing of anything which has to do with graphics, and also depends heavily on driver/GL implementation behavior, is going to be very useful.

> Ah, I had mistakenly assumed that double buffering automatically implied vsync, which I discovered through empirical testing with GLX FreeGLUT applications.

It's not even a matter of "X11/GLX behaves like this, win32 behaves like that". It varies between implementations. Nvidia on GNU/Linux allows the user to force vsync on/off from global settings or an environment variable, and defaults to vsync off. AMD (and probably all MESA drivers) default to vsync on, and also heeds another environment variable. SGI GL on IRIX always vsyncs with no way for an application to disable it, but it can be disabled by setting an environment variable.

drewwoods:
> It's not even a matter of "X11/GLX behaves like this, win32 behaves like that". It varies between implementations. Nvidia on GNU/Linux allows the user to force vsync on/off from global settings or an environment variable, and defaults to vsync off. AMD (and probably all MESA drivers) default to vsync on, and also heeds another environment variable. SGI GL on IRIX always vsyncs with no way for an application to disable it, but it can be disabled by setting an environment variable.

Thanks for the clarification

> I don't think it's worthwhile if it will only have a single regression test :) It might be useful if we were to create a whole suite of tests, and have them run automatically like the build testing we currently have as github actions, but that's a lot of work. Also I'm not convinced that automated testing of anything which has to do with graphics, and also depends heavily on driver/GL implementation behavior, is going to be very useful.

True

Given that I'm closing this PR

Issue #200, opened 2025-03-07
Unify FreeGLUT documentation sources with GLUT man pages


drewwoods:
## Background

@markkilgard recently gave permission to use original GLUT man pages to improve the FreeGLUT documentation: https://sourceforge.net/p/freeglut/mailman/message/59152910/

We currently have three sources of incomplete documentation in different formats:

1. **FreeGLUT Website Documentation**: https://freeglut.sourceforge.net/docs/api.php
   - Incomplete but focuses on FreeGLUT-specific functionality and differences from GLUT
   - Currently in PHP format

2. **API.md in the GitHub repository**: https://github.com/freeglut/freeglut/blob/master/doc/api.md
   - A subset of the website content
   - Markdown format

3. **Original GLUT man pages** (newly available)
   - Complete coverage of original GLUT functionality
   - Traditional Unix man page format
   - Lacks FreeGLUT-specific extensions and changes

## Problem

We ideally need to consolidate these sources into a unified documentation system that:
- Leverages the completeness of the GLUT man pages
- Preserves FreeGLUT-specific information and extensions
- Can be maintained from a single source
- Can be published in multiple formats (man pages, website, markdown)

## Proposed Solution: Documentation Maintenance Workflow

Adopt a single markup-based documentation source that can be transformed into the required outputs: man pages, PHP webpages, and Markdown. This approach ensures consistency and reduces maintenance complexity.

### Workflow Diagram
A single source document is processed by a transformation tool to generate the desired outputs:

```
              ┌──────────┐          
              │          ├────► man pages
              │          │          
 doc source ─►│transform ├────► PHP webpages
              │          │          
              │          ├────► Markdown (API.md)
              └──────────┘                                                
```

### Source Format Options
The "doc source" could be written in one of the target formats (since all are markup-based) and transformed into the others. 
For example, Markdown pages as source OR Man pages as source:
```                                                                                       
               ┌──────────┐                                     ┌──────────┐                                     
               │          ├────► man pages                      │          ├────► PHP webpages                   
     md   ────►│transform │                    OR     man  ────►│transform │                                     
               │          ├────► PHP webpages                   │          ├────► Markdown                       
               └──────────┘                                     └──────────┘                                                                                                                                                
```                                                                                                                           

**Alternatively** Other markup like reStructuredText or AsciiDoc, which are a neutral format could be used if it better supports transformation into all three outputs.

### Steps to Implement
1. **Choose a Source Format**:
[...]

PR #205, opened 2025-03-14
Add GLUT man pages from Mark Kilgard's original repository

drewwoods:
Import official GLUT man pages from Mark Kilgard's repository

- Source: https://github.com/markkilgard/glut/tree/master/man
- Imported with permission; copyright retained by original author.

CMake changes:
- Added `FREEGLUT_BUILD_MAN_PAGES` option (default: OFF).
- Installs pages to standard `man3` directory on Unix systems.

Man page modifications (verbatim copy except):
- Removed trailing whitespace and `^M` carriage returns.

See Issue #200 for details.


CMakeLists.txt
```diff
@@ -840,3 +840,14 @@ INSTALL(
     DESTINATION ${FREEGLUT_INSTALL_CMAKECONFIGDIR}
     COMPONENT Devel
 )
+
+# Install GLUT man pages on Unix-like systems
+IF(UNIX AND FREEGLUT_BUILD_MAN_PAGES)
```
drewwoods, line 845:
```suggestion
IF(UNIX AND FREEGLUT_INSTALL_MAN_PAGES)
```

now thinking about it `FREEGLUT_INSTALL_MAN_PAGES` is probably more appropriate

jtsiomb (approved):
Thank you.

One thing we need to pay attention to however, to maximise the usefulness of these man pages, is to make it clear they are provided under a free license. Otherwise (some) distributions won't touch them. Since Mark mentioned in his email that we are free to re-license under any free license as long as it's not the GPL, we should do this explicitly by adding a license file in the man directory (with Mark's copyright and the text of the license), and also change every manpage's header to reference the license right after the copyright line(s).

We can do this now, or I could merge this and change it later. Whichever you prefer. Let me know.

I think for the manpages it makes sense to either use the same license as the whole of freeglut, or use one of the creative commons licenses. I lean towards the first just to make things simpler.

drewwoods:
@jtsiomb okay I believe I did what you requested:
 - new [man/LICENSE](https://github.com/freeglut/freeglut/blob/03ce6d92c5aa63917b4f648ec6e6848f6593ebe7/man/LICENSE) file containing the FreeGLUT LICENSE wording
 - Added language to each man page to reference this license file
 - The [script](https://github.com/freeglut/freeglut/blob/69d993b6b223ceb28ddba9882e0f4da69b962dd4/src/util/update_man_license.sh) used to insert the license, which doesn't need to be part of the commit

jtsiomb:
Thank you

PR #209, opened 2025-03-22
Fix some tabs and indentation

tanksdude:
Fixes tabs and indentation in doc files, plus a few minor cases in the source files. Didn't touch every source file for tabs because there's way too many.


jtsiomb:
Thanks, but I'd rather not pull whitespace patches. If some parts of the code are inconsistent, they can stay that way until replaced by a functional change. Or maybe when some functional change is done to adjacent lines, it might do a slight reformatting of its immediate vicinity to make it more coherent.

tanksdude:
Got it 👍

PR #217, opened 2025-04-10
Man page and LICENSE updates

drewwoods:
Update glutGet man page with FreeGLUT enhancements and license

- Added copyright for The FreeGLUT Project Authors to LICENSE and glutGet.man
- Updated liability clause in LICENSE to include all copyright holders
- Enhanced glutGet.man with new state queries from FreeGLUT
- update .TH title header arguments for consistency and clarity
- Extend the update_man_license.sh script to perform these .TH header updates automatically.

See Issue #200


man/glutGet.man
```diff
@@ -143,12 +161,16 @@ The height value of the initial window size.
...
-The window system dependent format ID for the current layer of the
-current window.  On X11 GLUT implementations, this is the X visual ID.
-On Win32 GLUT implementations, this is the Win32 Pixel Format
-Descriptor number.  This value is returned for debugging, benchmarking,
-and testing ease.
+.B GLUT_FULL_SCREEN
+One if the window is currently in fullscreen mode, zero otherwise.
+.TP 8
+.B GLUT_INIT_STATE
+One if GLUT has been initialized through a call to glutInit, zero otherwise.
```
drewwoods, line 168: I’m referring to FreeGLUT as GLUT here to maintain consistency with the rest of the man page. I’m not sure if it would be better to replace all instances of GLUT with FreeGLUT in the description (excluding the enums, of course)?

man/glutGet.man
```diff
@@ -1,5 +1,6 @@
 .\"
 .\" Copyright (c) Mark J. Kilgard, 1996.
+.\" Modifications Copyright (c) The FreeGLUT contributors.
```
drewwoods, line 3: I’m uncertain about the proper capitalization of FreeGLUT. Should it be Freeglut, freeglut, or FreeGLUT?

jtsiomb: the preferred capitalization is all lowercase "freeglut".

jtsiomb: also I don't think the "Modifications"  prefix is necessary. Just a bunch of copyright lines is what I usually see in such cases. That later copyright lines are about modifications to the earlier content is implied.

drewwoods: great, made those modifications

drewwoods:
https://github.com/freeglut/freeglut/blob/f4f846f5dc390a34817d12e41e71dc497e1970ea/man/glutGet.man#L7

One more clarification point. What should the arguments to Title Header (`.TH`) be? I found [references](https://man.cx/groff_man(7)) to how the arguments are formatted, but I’m not sure of the convention on their meaning.

`.TH topic section [footer-middle] [footer-inside] [header-middle]`

Specifically should we modify:
1. "3.8" argument for `footer-middle`? Here it is used as the GLUT version number. So should we modify this to freeglut's version? Alternatively, other man pages use this as the date last modified
2. “GLUT” used for `footer-inside` and `header-middle`? Would "freeglut" be more appropriate?

jtsiomb:
> Specifically should we modify:
> 
>     1. "3.8" argument for `footer-middle`? Here it is used as the GLUT version number. So should we modify this to freeglut's version? Alternatively, other man pages use this as the date last modified

I vote for date last modified, because this is bound to become stale and just keep an arbitrary version number for ever.

>     2. “GLUT” used for `footer-inside` and `header-middle`? Would "freeglut" be more appropriate?

probably, yes.

PR #222, opened 2025-04-21
init: Improve display string parsing

drewwoods:
FreeGLUT’s glutInitDisplayString support currently parses only display modes, not criteria as specified in the full GLUT spec. This PR begins addressing that limitation.

This includes the following commits:

1. **init: fix "aux" token in display string parser**
     - Rename "aux" to "auxbufs" in the Tokens[] array for consistency with the
documented token name.

2. **init: add "=N"”parsing for `samples` and `auxbufs`**
     - Introduce `parseCriteria()` helper (ported from classic GLUT) plus
       `Comparison`/`Criterion` types to interpret numeric criteria.
     - `glutInitDisplayString()` now recognises the forms `samples=N` and
       `auxbufs=N`, setting `fgState.SampleNumber` and
       `fgState.AuxiliaryBufferNumber` respectively.

3. **Cocoa: honor auxbufs=N and samples=N when building the NSOpenGL pixel‑format**

### DIFF
--- a/src/fg_init.c
+++ b/src/fg_init.c
@@ -323,6 +323,86 @@ void fgDeinitialize( void )
     fgState.Initialised = GL_FALSE;
 }
 
+/* -- HELPER FUNCTIONS -------------------------------------------------- */
+
+typedef enum {
+  NONE, EQ, NEQ, LTE, GTE, GT, LT, MIN, INVALID
+} Comparison;
+
+typedef struct _Criterion {
+  Comparison comparison;
+  int value;
+} Criterion;
+
+/* Parse a display string token and return a Criterion struct.
+ * Based on the original GLUT implementation:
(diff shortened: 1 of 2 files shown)


src/fg_init.c
```diff
@@ -444,12 +524,13 @@ static char* Tokens[] =
     "xstaticgray", "xgrayscale", "xstaticcolor", "xpseudocolor",
     "xtruecolor", "xdirectcolor",
     "xstaticgrey", "xgreyscale", "xstaticcolour", "xpseudocolour",
-    "xtruecolour", "xdirectcolour", "borderless", "aux"
+    "xtruecolour", "xdirectcolour", "borderless", "auxbufs"
```
jtsiomb, line 527: We should accept both "auxbufs" and "aux", to maintain backwards compatibility with earlier freeglut versions.

drewwoods: good point

PR #232, opened 2025-09-15
demos: add accumulation buffer motion blur example

drewwoods:
Add a minimal interactive demo showing accumulation buffer usage for motion blur effects. Features a rotating line/cross with configurable parameters including rotation speed, line dimensions, accumulation passes, and rendering options.

### DIFF
--- a/progs/demos/accum/accum.c
+++ b/progs/demos/accum/accum.c
@@ -1,0 +1,220 @@
+/*
+ * accum.c - Accumulation Buffer Motion Blur Example
+ * Written by Andrew Woods <drew.woods at gmail.com>
+ *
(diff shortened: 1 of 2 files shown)


progs/demos/accum/accum.c
```diff
@@ -0,0 +1,219 @@
...
+ * Esc:          Exit the program
+ */
+
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <GL/freeglut.h>
+
+// Application state
```
jtsiomb, line 29: please make all comments in C files C89-compatible

jtsiomb (changes requested):

PR #238, opened 2025-10-29
x11: ensure iconified windows are properly restored via _NET_ACTIVE_WINDOW

drewwoods:
When testing #237 on X11, I found that windows minimized via glutIconifyWindow() did not reappear when glutShowWindow() was called (specifically  [here](https://github.com/freeglut/freeglut/pull/237/files#diff-5833c849769ce6bb9ac7aca781e39af879e3cbcff7aa69be2d24fb70601fbce7R263-R275)).

Some GNOME-based window managers (such as Cinnamon and Mutter) ignore plain XMapWindow requests for iconified windows, requiring an explicit _NET_ACTIVE_WINDOW client message to reactivate them.

This patch adds that missing XEvent message, ensuring that glutShowWindow() correctly restores minimized windows.

Note: Tested on linux mint 22 (ubuntu 24.04 based) with Cinnamon

### DIFF
--- a/src/x11/fg_window_x11.c
+++ b/src/x11/fg_window_x11.c
@@ -493,7 +493,26 @@ void fgPlatformCloseWindow( SFG_Window* window )
  */
 void fgPlatformShowWindow( SFG_Window *window )
 {
+    XEvent xev;
+    long evmask = SubstructureRedirectMask | SubstructureNotifyMask;
+
     XMapWindow( fgDisplay.pDisplay.Display, window->Window.Handle );
+
+    /* Send _NET_ACTIVE_WINDOW client message to the window manager to ensure iconified windows
+     * are restored.  This is required for GNOME flavored window managers Mutter, Cinnamon, etc.
+     */
+    memset( &xev, 0, sizeof( xev ) );
+    xev.type                 = ClientMessage;
+    xev.xclient.window       = window->Window.Handle;
+    xev.xclient.message_type = XInternAtom( fgDisplay.pDisplay.Display, "_NET_ACTIVE_WINDOW", False );
+    xev.xclient.format       = 32;
+    xev.xclient.data.l[0]    = 1;           /* source indication: 1 = application */
+    xev.xclient.data.l[1]    = CurrentTime; /* timestamp */
+    xev.xclient.data.l[2]    = 0;           /* Active window (0 = current window) */
+    if ( !XSendEvent( fgDisplay.pDisplay.Display, fgDisplay.pDisplay.RootWindow, 0, evmask, &xev ) ) {
+        fgWarning( "Failed to send _NET_ACTIVE_WINDOW message" );
+    }
+
     XFlush( fgDisplay.pDisplay.Display ); /* XXX Shouldn't need this */
 }
 


drewwoods:
@jtsiomb would you happen to have access to any other X11 window managers? Specifically legacy SGI stuff.

All my *nix boxes have GNOME based ones, but I can install some others and test sometime.

I would prefer not to land this before its more extensively tested.

jtsiomb:
Yes, I can test on the SGI 4Dwm soon. I can also test on my main GNU/Linux machine where I'm running fvwm.

jtsiomb:
@drewwoods restoring the iconified window works on both window managers I tested (4Dwm, fvwm). On 4Dwm some windows are created with their titlebar off-screen and they stay there:

fgwindows

drewwoods:
I just noticed that you have a non-zero `GLUT_WINDOW_BORDER_WIDTH` in the output that you did share.   I always tested it on machines where it was 0, so when it was used in a calculation, it had no effect, so I may be excluding it from some places. I assumed the `GLUT_WINDOW_BORDER_WIDTH` was just affecting the window width, but it might affect the hight too. 

https://github.com/freeglut/freeglut/blob/f292cfe812ab054cd5349588f66e7ed2f38c5a1c/progs/demos/windows/windows.c#L74-L100

So the `GLUT_WINDOW_BORDER_WIDTH` might need to be included in determining the actual Y position and height of the window. 

```diff
diff --git a/progs/demos/windows/windows.c b/progs/demos/windows/windows.c
index 0ec1ad91..15915dfd 100644
--- a/progs/demos/windows/windows.c
+++ b/progs/demos/windows/windows.c
@@ -86,7 +86,7 @@ void get_geom( void )
      *
      * This differs from the original GLUT behavior. See docs/api.md#conventions
      */
-    winGeom[win].y -= glutGet( GLUT_WINDOW_BORDER_HEIGHT );
+    winGeom[win].y -= glutGet( GLUT_WINDOW_BORDER_HEIGHT ) + glutGet( GLUT_WINDOW_BORDER_WIDTH );
 
     /*
      * In freeglut and the original GLUT, glutGet( GLUT_WINDOW_WIDTH/HEIGHT )
@@ -95,7 +95,7 @@ void get_geom( void )
      * Unlike GLUT, freeglut provides GLUT_WINDOW_BORDER_WIDTH/HEIGHT
      */
     winGeom[win].w += 2 * glutGet( GLUT_WINDOW_BORDER_WIDTH );
-    winGeom[win].h += glutGet( GLUT_WINDOW_BORDER_HEIGHT );
+    winGeom[win].h += 2 * glutGet( GLUT_WINDOW_BORDER_WIDTH ) + glutGet( GLUT_WINDOW_BORDER_HEIGHT );
 #endif
 }
```

drewwoods:
> Yeah there is horizontal movement, and the teapot window resizes continuously, here's a short video.
> 
>  sgi_fgwindows.mp4 
>

Awesome, thanks for the video, although sad to see the windows' nerfed bouncing ability 😅. 

I suspect it has something to do with `if (Y <= 0) Ydir = 1` condition not being hit. But this isn't the PR to discuss that.

Just to confirm, is there an issue with restoring the iconified windows without this PR in 4Dwm and fvwm? I’m wondering if this is because of the evolution of XWindow managers or a latent bug in freeglut?

PR #192, opened 2025-01-23
macOS demos: All demos successfully build on macOS

drewwoods:
Notes:
- OpenGL compatibility mode on macOS uses different function prototype typedefs, requiring redefinition

- While macOS OpenGL compatibility mode is limited to OpenGL 2.1, it provides extensions approximating OpenGL 3.3 functionality. However, these extensions use different naming conventions

<details>
<summary>Build Errors Master</summary>

```
> make -j > /dev/null
ld: warning: search path '/usr/local/opt/gettext/lib' not found
/Users/drew/src/freeglut/progs/demos/smooth_opengl3/smooth_opengl3.c:132:1: error: unknown type name 'PFNGLGENBUFFERSPROC'
  132 | PFNGLGENBUFFERSPROC gl_GenBuffers;
      | ^
/Users/drew/src/freeglut/progs/demos/smooth_opengl3/smooth_opengl3.c:133:1: error: unknown type name 'PFNGLBINDBUFFERPROC'
  133 | PFNGLBINDBUFFERPROC gl_BindBuffer;
      | ^
/Users/drew/src/freeglut/progs/demos/shapes/shapes.c:115:1: error: unknown type name 'PFNGLCREATESHADERPROC'
  115 | PFNGLCREATESHADERPROC gl_CreateShader;
      | ^
/Users/drew/src/freeglut/progs/demos/smooth_opengl3/smooth_opengl3.c:134:1: error: unknown type name 'PFNGLBUFFERDATAPROC'
  134 | PFNGLBUFFERDATAPROC gl_BufferData;
      | ^
/Users/drew/src/freeglut/progs/demos/shapes/shapes.c:116:1: error: unknown type name 'PFNGLSHADERSOURCEPROC'
  116 | PFNGLSHADERSOURCEPROC gl_ShaderSource;
      | ^
/Users/drew/src/freeglut/progs/demos/shapes/shapes.c:117:1: error: unknown type name 'PFNGLCOMPILESHADERPROC'
  117 | PFNGLCOMPILESHADERPROC gl_CompileShader;
      | ^
ld: warning: ignoring duplicate libraries: '-lm'
ld: warning: search path '/usr/local/opt/gettext/lib' not found
/Users/drew/src/freeglut/progs/demos/smooth_opengl3/smooth_opengl3.c:135:1: error: unknown type name 'PFNGLCREATESHADERPROC'
  135 | PFNGLCREATESHADERPROC gl_CreateShader;
      | ^
[...]

PR #253, opened 2025-11-13
docs: clarify that demo builds work on macOS and standardize naming

drewwoods:
- Updated README.macos to note that building demos is supported (see 27f8d8c)
- Standardized capitalization of "macOS" throughout
- See also #192

### DIFF
--- a/README.macos
+++ b/README.macos
@@ -1,10 +1,10 @@
-Freeglut on MacOS X
-===================
+Freeglut on macOS
+=================
 
-Freeglut on macos can be compiled to either rely on an X server and GLX, or use
+Freeglut on macOS can be compiled to either rely on an X server and GLX, or use
 the native Cocoa and OpenGL frameworks. Freeglut versions before 3.8.0 only
 supported X11/GLX, and it's the most complete and well-tested version, but it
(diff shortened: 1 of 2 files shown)


jtsiomb:
Hmmmmm I don't know. I think I like "MacOS X" better. "macOS" feels unbalanced, and also I find dropping the X confusing, what would you call MacOS 9 then?

drewwoods:
Yeah, I get what you mean, Apple’s naming history hasn’t exactly been a model of consistency. The whole Mac OS 9 --> Mac OS X --> OS X --> macOS progression has been confusing even for people who follow the platform closely. The recent jump from macOS 15 straight to macOS 26 doesn’t help either.

Since Apple switched to the lowercase macOS branding back in 2016, most projects seem to have converged on that, and it helps keep CI configs, docs, and filenames aligned. The goal here was simply to signal that freeglut supports modern macOS, using the older naming might accidentally suggest it targets legacy versions.

But your point about how this clashes with the old Mac OS 9 naming is totally fair. If you think it would help, we could add a brief note explaining the terminology shift the first time it comes up.

PR #284, opened 2026-06-24
cocoa: implement glutSetCursor for Cocoa backend

drewwoods:
How it works: rather than forcing the cursor globally, it uses AppKit's cursor-rect mechanism so the cursor is per-window and only applies while the pointer is over that window's view — matching GLUT semantics on X11/Windows.

- src/cocoa/fg_internal_cocoa.h — added a Cursor field (NSCursor *) to CocoaPlatformContext; nil represents GLUT_CURSOR_INHERIT.
- src/cocoa/fg_cursor_cocoa.m — fgPlatformSetCursor maps the GLUT cursor ID to an NSCursor, stores it on the window, and invalidates cursor rects for that window and (recursively) its subwindows so inheriting children update too. fghCocoaEffectiveCursor walks up the parent chain to resolve inheritance, same as X11's XUndefineCursor behavior.
- src/cocoa/fg_window_cocoa.m — fgOpenGLView overrides resetCursorRects to apply the resolved cursor over its bounds, and fgPlatformCloseWindow releases the stored cursor.

Mapping notes, since macOS's public cursor set doesn't cover everything GLUT defines:

- GLUT_CURSOR_NONE uses a transparent 16×16 cursor image (the per-window equivalent of X11's empty pixmap cursor — NSCursor hide would have been global).
- The four corner-resize cursors use the frameResizeCursorFromPosition:inDirections: API on macOS 15+, falling back to crosshair on older systems/SDKs.
- HELP, CYCLE, and WAIT fall back to the arrow (no public equivalents; macOS shows its own beachball when busy), SPRAY and FULL_CROSSHAIR demote to crosshair (X11 demotes FULL_CROSSHAIR the same way), and both arrows map to the standard arrow since macOS has no mirrored variant. Unknown IDs hit fgError, matching X11.

NOTE: written mainly using Claude Opus 4.8

TESTING: This cursor implementation has been used in private repositories and works as expected. As mentioned above, not all cursors are natively supported on macOS.

### DIFF
--- a/src/cocoa/fg_cursor_cocoa.m
+++ b/src/cocoa/fg_cursor_cocoa.m
@@ -62,9 +62,137 @@ void fghPlatformGetCursorPos( const SFG_Window *window, GLboolean client, SFG_XY
     mouse_pos->Use = GL_TRUE;
 }
 
+/*
+ * Cursor handling works through AppKit cursor rects: fgPlatformSetCursor
+ * stores the NSCursor for the window in pContext.Cursor, and the window's
+ * fgOpenGLView applies it from -resetCursorRects whenever the pointer is
+ * inside the view. A nil cursor means GLUT_CURSOR_INHERIT.
+ */
+
+/* A fully transparent cursor, for GLUT_CURSOR_NONE */
+static NSCursor *fghEmptyCursor( void )
+{
+    static NSCursor *empty = nil;
+
+    if ( !empty ) {
+        NSImage *blank = [[NSImage alloc] initWithSize:NSMakeSize( 16, 16 )];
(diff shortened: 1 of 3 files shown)


jtsiomb:
Make sure to be extra vigilant in reading, understanding, and reviewing any code partly written by an LLM. They tend to produce insane things from time to time.

drewwoods:
> Make sure to be extra vigilant in reading, understanding, and reviewing any code partly written by an LLM. They tend to produce insane things from time to time.

Ye, this one is mainly plumbing, so was an easy review.  That said, frontier models are becoming remarkably resourceful. A follow-on PR, once reviewed properly, is quite impressive. Opus used the appropriate tooling to one-shot generate some of the missing macOS cursors.

PR #285, opened 2026-06-24
CI: pin Windows MSVC runner to 2022

drewwoods:
Addresses MSVC [build failures](https://github.com/freeglut/freeglut/actions/runs/28124589745/job/83284999800)

### DIFF
--- a/.github/workflows/build_win_msvc.yml
+++ b/.github/workflows/build_win_msvc.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   build:
 
-    runs-on: windows-latest
+    runs-on: windows-2022
 
     steps:
     - uses: actions/checkout@v3


jtsiomb:
I'll merge for now, but if you have any clue why it fails on latest, I'd rather we fixed it so that it doesn't break again when they remove that specific runner.

dcnieho:
It appears that more generally, the build/CI config can use some updates.
e.g. ***@***.*** is several versions old, and see the other
warnings about outdated things being used. Anyone want to take on that task?

github has a nice dependabot that keeps track of such outdated things and
automatically makes pull requests to update. I'll have a look if we can
configure that for this repo (it of course remains an option to ignore the
pull request).

Cheers,
Dee

drewwoods:
> I'll merge for now, but if you have any clue why it fails on latest, I'd rather we fixed it so that it doesn't break again when they remove that specific runner.

good point, addressed in #286

PR #286, opened 2026-06-25
CI: use latest Windows runner without explicit MSVC generator

drewwoods:
The previous commit, 9cf2f8cc177ebd74160a991d1a6408c122fdf466, switched to the windows-2022 runner to match the explicit Visual Studio 2022 generator.

A less brittle solution is to use the latest Windows runner and let CMake select the appropriate generator.

See also #285


jtsiomb:
Thanks for looking into it, I didn't remember that we asked for a specific MSVC version previously.

PR #288, opened 2026-07-16
add glutGet64 and high resolution nanosecond time support

lndpj:
- [x] `uint64_t glutGet64(GLenum eWhat)`
- [x] support vsync demo
- [x] `GLUT_ELAPSED_NS_TIME`

**NsTime**

- [x] x11
- [x] mswin
- [ ] android
- [ ] blackberry
- [ ] cocoa
- [ ] egl
- [ ] ogc
- [ ] wayland

### DIFF
--- a/include/GL/freeglut_std.h
+++ b/include/GL/freeglut_std.h
@@ -309,7 +309,9 @@
 #define  GLUT_INIT_WINDOW_HEIGHT            0x01F7
 #define  GLUT_INIT_DISPLAY_MODE             0x01F8
 #define  GLUT_ELAPSED_TIME                  0x02BC
-#define  GLUT_ELAPSED_NS_TIME               0x02BD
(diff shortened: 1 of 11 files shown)


jtsiomb:
I appreciate the effort, but:

1. have you noticed the discussion about this in the issue you opened?
2. have you noticed how many unrelated things you changed besides adding "glutGet64 and high resolution nanosecond time support"?
3. when it comes to adding new freeglut API entry points, there's got to be a lot more discussion about it, before deciding if we're going to add any, how it's going to work, and what's it going to be called. That's because from that point on it will be set in stone and will have to be maintained for eternity. So discuss first, send pull requests later.
4. This whole changeset has no consideration for cross-platform compatibility. So even if all of the above were not issues, it would still need heavy rework before being accepted.

PR #291, opened 2026-08-10
demos: new freeglut cursors demo

drewwoods:
Demonstrates freeglut resize cursors with a resizable rectangle.  Resize the rectangle by dragging its edges and corners, or click its centre to spin it!

Whee:
cursor-demo

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -776,6 +776,7 @@ ADD_DEMO(accum           progs/demos/accum/accum.c)
 ADD_DEMO(windows         progs/demos/windows/windows.c)
 ADD_DEMO(vsync           progs/demos/vsync/vsync.c)
 ADD_DEMO(fonts           progs/demos/fonts/fonts.c)
+ADD_DEMO(cursors         progs/demos/cursors/cursors.c)
 
 #######################
 # INSTALLATION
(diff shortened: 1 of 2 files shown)


jtsiomb:
This is extremely cute, but I don't think it makes sense to have a whole demo just to demonstrate *two* of the possible cursors. Wouldn't it be better (albeit more boring), to have a menu with all the possible GLUT cursors, and being able to select any one of them to set?

drewwoods:
FWIW the demo uses 9 GLUT cursors, but granted the GIF was advertising the spin more than the cursors :P .  I understand the pushback, since this doesn't quite fit the theme of the other demos, and the spin is probably a step too far rather than demonstrating the cursors themselves.
https://github.com/freeglut/freeglut/blob/c5e4239f46185af5d8322b2321d1ca8ea7f8c181/progs/demos/cursors/cursors.c#L90-L100

Would you be happy if we dropped the spin? This would reduce the code by ~40% at the cost of the GLUT_CURSOR_CYCLE, so it would still demonstrate 8 cursors.

dcnieho:
I'd say lets keep the spin, but maybe also demonstrate the other cursors
not related to resizing/spin? You could show those by hovering over
specific other areas?

jtsiomb:
Ok then, whee it is :)

Let's pull that in, and we can add more cursors to it later at some point.

PR #294, opened 2026-08-12
cocoa: avoid client pixel-store stack for bitmap fonts

drewwoods:
Apple's OpenGL compatibility implementation handles explicit pixel-store queries and restores substantially more cheaply than `glPushClientAttrib(GL_CLIENT_PIXEL_STORE_BIT)`.

See [bitmap-font-state-bench](https://github.com/drewwoods/freeglut/tree/bitmap-font-state-bench/progs/demos/bitmap_bench) for the benchmark code and detailed results.  Note the benchmark and report was generated by Claude Opus 5.0, but **verified**.

This was noticed when rendering ~1000 glyphs took ~2ms when using `glutBitmapChar()`

The overhead is mainly due to the state management performed per glyph. `glutBitmapString()` amortizes this overhead, so for typical-length strings the runtime is dominated by bitmap rasterization. 

Performance is trashed by larger glyphs for reasons unknown.

```
 Cost per glutBitmapCharacter (in situ), Apple GL / Cocoa:
                  clientattrib   getset
   8x13               1885 ns    492 ns    3.8x
   9x15               1950 ns    529 ns    3.7x
   tr10               1875 ns    459 ns    4.1x
   hel10              1889 ns    461 ns    4.1x
   hel12              1900 ns    492 ns    3.9x
   hel18             27400 ns  28000 ns    1.0x
   tr24              66500 ns  67600 ns    1.0x
```

### DIFF
--- a/src/fg_font.c
+++ b/src/fg_font.c
@@ -92,7 +92,12 @@ void FGAPIENTRY glutBitmapCharacter( void* fontID, int character )
     /* Find the character we want to draw */
     face = font->Characters[ character ];
 
-#ifdef GL_VERSION_1_1
+    /*
+     * Apple's OpenGL compatibility implementation handles the explicit
+     * pixel-store queries/restores below much more cheaply than its client
+     * attribute stack.  Only the six values changed here need preserving.
+     */
+#if defined(GL_VERSION_1_1) && !TARGET_HOST_MACOS_COCOA
     glPushClientAttrib( GL_CLIENT_PIXEL_STORE_BIT );
 #else
 	{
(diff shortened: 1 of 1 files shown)


drewwoods:
I'll rerun these -- the micro benchmarks are giving me state management overheads, not per glyph costs with the two different state management techniques.  OpenGL state management overheads are very operation specific, so testing in isolation is kind of meaningless

...Done

src/fg_font.c
```diff
@@ -92,7 +92,25 @@ void FGAPIENTRY glutBitmapCharacter( void* fontID, int character )
...
-#ifdef GL_VERSION_1_1
+    /*
+     * Apple's OpenGL compatibility implementation handles the explicit
+     * pixel-store queries/restores below much more cheaply than its client
+     * attribute stack.  Verified this also helps the XQuartz path on macOS.
+     *
+     * See https://github.com/drewwoods/freeglut/tree/bitmap-font-state-bench
+     * for benchmark code and detailed results.
+     *
+     * Cost per glutBitmapCharacter (in situ), Apple GL / Cocoa:
```
jtsiomb: Including the actual benchmarks in this comment is I think a bit too much. Put them in the commit message if you want the documentation to persist, but keep the comment short: apple's blah blah sucks, so we won't use push attrib

jtsiomb (changes requested):

src/fg_font.c
drewwoods: Lol, ye it does suck - MESA / Nvidia presumably use dirty flags on the push/pop attrib paths

PR #297, opened 2026-08-27
demos: add mouse + wheel callback counter

drewwoods:
Add a simple mouse input demo that reports wheel and mouse button callbacks, including direction, button state, coordinates, and totals.

Allow toggling the wheel callback at runtime to exercise freeglut's fallback handling of wheel events through the mouse callback.

This demo was generated by AI and I have no particular fondness of it, a little too utilitarian for my tastes :stuck_out_tongue_winking_eye:. That said, it does cover this use case, which none of the other demos do.

This was used to verify
- #296


jtsiomb:
The code feels very clunky, and I'm not sure we need a dedicated mousewheel demo. We're not trying to assemble a test suite here, these are just demos.

Also what's that about multiple mousewheels? sure the mousewheel gets mapped to button 4 and 5, but beyond that it's just buttons afaik.

drewwoods:
Yeah, totally happy to kill this. Something similar can be regenerated with a simple prompt and somewhat hauntingly, that’s only going to get better.

